PR #21, opened 2020-10-19
Replace section "Overview" by "Getting started"

### LINKED ISSUE #20: Set up basic structure of docs
Adapt structure according to [discussion in telco](https://redmine.iks.cs.ovgu.de/oe/projects/ego-n/wiki/05102020_Powertag_DP_Workflow#Dokumentation)

gplssm:
Fix #20


gplssm:
I'm done so far with the basic structure. Remaining tasks for filling the docs with content will be created afterwards. Each one for a single task.

### Open questions

- We could some merge "getting started", "usage" and "workflow" (at least down to two sections). Opinions?
- Shall we keep and maintain the "Authors" list?

gplssm:
...meanwhile, I gonna check why tests on travis fails (maybe there aren't any)

gnn:
As you can see, I pushed a fix, which hopefully deals fixes the Travis build. In the meantime, here are few tips:

  - You can always run he whole build by running `tox` in the repository. You need to have all the supported Python versions installed for this though.
  - If you only have one of the supported Python versions, e.g. Python 3.8, installed, you can test the build for this version by running `tox -e py38-cover` for testing the build with coverage enabled and `tox -e py38-nocov` for testing the build without coverage.
  - Regardless of your Python version, you can always run `tox -e check` to run various checks, e.g. import ordering, PEP8 violations, validity of links in the decumentation, etc. against the code. These are run by Travis lead to a failing build if one of them fails.
  - You can also build the documentation locally by running `tox -e docs` regardless of your installed Python version. Again, this is ran by a Travis build, so if the documentation fails to build, the Travis build fails too. At least think so.

All of this should probably go into the package documentation. Feel free to copy it, or point me to a location where you would like to read, and I'll try include it there. In the long run, I'll probably put it into the "README.rst".

Last but not least: "Fix #20" (or rather "Fixes #20") is a prime example of something which would be a perfect commit message for a merge commit. Because when all the tests pass, you know that you can merge and that your commit actually fixes the issu. :)

gnn (commented):
Classic case of a merge commit that can be avoided by using `git pull -r`, which I would definitely recommend in moste cases.

gnn:
Lookint into the Travis failure, I see that only the tests fail, for reasons which this PR can't do anything about. The good news is that  already fixed this on another branch, so once that gets merged, these failures will be gone.

gplssm:
> Looking into the Travis failure, I see that only the tests fail, for reasons which this PR can't do anything about. The good news is that already fixed this on another branch, so once that gets merged, these failures will be gone.

Good to hear! So, @IlkaCu just have to take a look on the structure of the documentation and can approve then. But, it's an pragmatic exception we're taking here :wink:

IlkaCu (approved):
The basic structure looks good to me.

PR #28, opened 2020-10-20
Fix #26

### LINKED ISSUE #26: RTD badge in README points to non-existent page
Right now, clicking on the RTD badge redirects to https://readthedocs.org/projects/eGon-data. This page does not exist.
I propose to use https://egon-data.readthedocs.io instead.

gplssm:


gnn:
Commits both look great. If you don't mind, I'd like to rewrite the commit messages, especially the first, as that one doesn't convey any information without the GH repository, e.g. when doing a `git log`.

Issue #18, opened 2020-10-13
Provide a local postgres database with docker


gplssm:
Use docker compose to spin up a temporary database for data processing with SQL.

I already did this in other repos. For the first try, simply use

- docker-compose.yml
- Dockerfile.postgis
- entrypoints/

from https://github.com/windnode/windnode_abw

gplssm:
Don't forget to write a short how-to-use in [pre-requisites](https://github.com/openego/eGon-data/blob/fa745ee7fd75b5e802d34db29ff37be7eb48aeb4/docs/getting_started.rst#L21)

PR #29, opened 2020-10-21
Start a docker container with a database via a DAG task

gnn:
This should address issue #18. It should be as "simple" as

  - running `egon-data serve`,
  - opening `localhost:8080` in a browser window,
  - turning the only available DAG `ON`.

You might have to trigger the DAG explicitly, after turning it `ON`. After a while `docker ps` should show that the container named `egon-data-local-database` is up.
I know that all of this should also be in the documentation. I'll copy it there, once I get to it.


gnn:
Closed because the PR got opened relative to the wrong state of `dev`.

Issue #7, opened 2020-10-09
Transfer zensus_vg250 scripts


IlkaCu:
Scripts related to preprocessing section zensus_vg250 need to be reviewed, adjusted and transfered to the new egon-data repo.

**Tasks**

- [x] Add data filtering for zensus with population > 0 and inside Germany
- [x] Add metadata for the resulting table of above, see [here](https://github.com/openego/data_processing/blob/5edca414212ccb4b1df6b046bf916bc2ebb7b4c6/preprocessing/zensus_vg250/ego_pp_destatis_zensus_metadata.sql)
- [x] Calculate population per municipality. Use the [open_ego data processing script](https://github.com/openego/data_processing/blob/5edca414212ccb4b1df6b046bf916bc2ebb7b4c6/preprocessing/vg250/ego_dp_structure_boundaries_vg250.sql#L450-L687). There might be a duplication with the first task
- [x] Add note to CHANGELOG

gplssm:
While I transferred vg250 import scripts, I've realized the is as well code that depends on the existence of census data. Since that is not the case right now, I move the remaining parts to here. So, in addition to the actual "zensus_vg250 scripts", integrate the following code as well:

https://github.com/openego/data_processing/blob/5edca414212ccb4b1df6b046bf916bc2ebb7b4c6/preprocessing/vg250/ego_dp_structure_boundaries_vg250.sql#L450-L687

Don't forget to update the table about required materialized views of vg250: https://github.com/openego/eGon-data/issues/68#issue-757060493

gplssm:
While working on this, I've realized that VG250 has partially invalid geometry

`SELECT bez, gen, ST_IsValid(geometry) FROM boundaries.vg250_sta;`

"Bundesrepublik"	"Deutschland"	false
"Bundesrepublik"	"Deutschland"	true
"Bundesrepublik"	"Deutschland"	true
"Bundesrepublik"	"Deutschland"	true
"Bundesrepublik"	"Deutschland"	true
"Bundesrepublik"	"Deutschland"	true
"Bundesrepublik"	"Deutschland"	true
"Bundesrepublik"	"Deutschland"	true
"--"	"Österreich (Bodensee)"	true
"--"	"Schweiz (Bodensee)"	true
"Bundesrepublik"	"Deutschland (Bodensee)"	true

[...]

PR #61, opened 2020-11-27
Features/#3 include vg 250

### LINKED ISSUE #3: Include VG_250
Include data on administrative districts and their boundaries from https://daten.gdz.bkg.bund.de/produkte/vg/vg250_ebenen_0101/
Scripts to import the data set can be derived from open_eGo, the latest data set from 2020 should be used - do not use folder 'aktuell' as this one is volatile. 
Come to a decision whether egon-data should provide this data on the OEP or not. 

### Tasks

- [x] Download data from server
- [x] Extract zip and import to DB
- [x] Create MVIEWS (`ego_pp_vg250_mview.sql` executed with `PostgresOperator`)
- [x] Add metadata
- [x] Check if content of `ego_dp_structure_boundaries_vg250.sql` is still needed
- [x] Re-check if everything from `ego_pp_vg250_import.sql` was included
- [x] Append tasks to pipeline
- [x] Append info to CHANGELOG

gplssm:
Integrate VG250 (Verwaltungsgebiete) in the data pipeline.

Fixes #3


gplssm:
I transferred VG250 import close to as done in @openego/data_processing. Thereby, I left out a bunch on materialized view creations which depend on the existence of census data. The task is transferred to here: https://github.com/openego/eGon-data/issues/7#issuecomment-738754694

Moreover, I would like to find out which of these views we actually do need. This will also be done later, see #68.

ClaraBuettner (approved):
I checked out to this branch and run airflow successfully. In addition, I verified that the desired datasets are included in the local database and the installation of egon-data works. I did not really looked at the commits but since the results are fine I approve this.

PR #135, opened 2021-03-04
Features/#9 migrate substation section

IlkaCu:
This PR integrates the abstraction of hvmv and ehv substations into the egon-data workflow and partly solves issues adressed in #9


ulfmueller (approved):
so far it worked very fine for me. I saw that in the old tables there is already a column 'otg_id' in those tables. I guess that column is added in a further script. So, so far I am happy with it :)

IlkaCu:
> I saw that in the old tables there is already a column 'otg_id' in those tables. I guess that column is added in a further script.
I will add this as soon as I continue with the assignment of the otg_id.

PR #144, opened 2021-03-09
Features/#124 integrate re potential areas

nesnoj:


ClaraBuettner:
Could you add a note to the [changelog](https://github.com/openego/eGon-data/blob/features/%23124-integrate-re-potential-areas/CHANGELOG.rst)? 
I tested the branch, it worked fine and the expected data is created. 
The tox testing seems to work differently for me, I don't get these `InterpreterNotFound` lines and it also checks every python script. The output is very long, I don't want to add it here. But I don't know much about these tests...

nesnoj:
> Could you add a note to the [changelog](https://github.com/openego/eGon-data/blob/features/%23124-integrate-re-potential-areas/CHANGELOG.rst)?
> I tested the branch, it worked fine and the expected data is created.
> The tox testing seems to work differently for me, I don't get these `InterpreterNotFound` lines and it also checks every python script. The output is very long, I don't want to add it here. But I don't know much about these tests...

Thanks for the quick review! :)
I updated the changelog.

I'll bring the tox question to the daily standup..

ClaraBuettner (approved):

nesnoj:
Thank you @ClaraBuettner !

PR #149, opened 2021-03-10
Features/#101 import heat demand data

### LINKED ISSUE #101: Heat Demands based on Peta5.0.1
Based on the residential and service-sector heat demand rasters of the Pan-European Atlas for 2015, future heat demand data for households and the service-sector have to be calculated for the German Census cells.

EvaWie:
Close #101
Based on the downloaded Europe-wide Peta5.0.1 heat demands per hectare for the year 2015, future heat demands per census cell within the boundaries of Germany are calculated.
The dataset version management (including numbering) and the referencing of underlying datasets in the meta data are open TODOs.


ClaraBuettner:
I'm wondering if we should rename the table, e.g. to `demand.egon_peta_heat` (egon because of our specific adjustments and peta because it is the main original source) What do you think?

ClaraBuettner (approved):

PR #154, opened 2021-03-11
Change port of default egon-data-local-database

### LINKED ISSUE #137: Change docker database port to less common one
Currently, when using a dockered database, the port 54321 is used. As this is a favorite of many (at least at RLI) and multiple docker database instances exist on the infrastructure I use, conflicts come up. I ran into the situation that the DAG failed because a wrong database was used :(

I suggest to change the port number to a less common one. 
@gnn are the any technical restrictions (length of the number) or are there any recommendations (preferred ranges)?

gplssm:
New, randomly chosen port is 59734

Fixes #137


gnn:
Since this got implemented in 2f21dae2f7f101e08e55513d503ace67ceb859ac, the PR can be closed and the branch can be deleted, right @gplssm?

PR #168, opened 2021-03-19
Features/#127 add hydro biomass plants e gon2035

### LINKED ISSUE #127: Integrate hydro and biomass power plants for eGon2035
Hydro and biomass power plants will be integrated based on data from MaStR and NEP 2021. 
Existing power plants from MaStR will be scaled linear to meet target values per federal state from NEP 2021 (prox2now expansion).

ClaraBuettner:
Closes #127 
This branch adds hydro and biomass plants to the power plant table for scenario eGon2035. 
There are still some things  missing:
- deal with plants without a geometry (<30kW, only small amount of biomass and hydro is affected by this) 
- assign voltage level (MaStR Netzverknüpfungspunkte)
- assign substation (depends on voltage level)

I'm not sure if it make sense to merge it right now or if we want to wait until those steps can be integrated.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -51,6 +51,8 @@ Added
   `#101 <https://github.com/openego/eGon-data/issues/101>`_
 * Demographic change integration
   `#47 <https://github.com/openego/eGon-data/issues/47>`_
+* Add hydro and biomass power plants eGon2035
+  `#127 <https://github.com/openego/eGon-data/issues/127>`_
 
 Changed
 -------
--- a/src/egon/data/airflow/dags/pipeline.py
+++ b/src/egon/data/airflow/dags/pipeline.py
@@ -112,19 +112,6 @@ with airflow.DAG(
     )
     vg250_clean_and_prepare >> demandregio_import
 
-    # Power plant setup
-    power_plant_tables = PythonOperator(
-        task_id="create-power-plant-tables",
-        python_callable=power_plants.create_tables
-    )
-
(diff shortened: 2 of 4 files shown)


IlkaCu:
> I'm not sure if it make sense to merge it right now or if we want to wait until those steps can be integrated.

I think both ways are reasonable, but it might be a bit nicer to integrate it now and catch up on the open tasks with a fresh features-branch as soon as the additional data/methods are available.

IlkaCu:
The following error occurred while executing import-power-plants: 
```
ERROR - (psycopg2.errors.UndefinedColumn) column "sources" of relation "egon_power_plants" does not exist
LINE 1: INSERT INTO supply.egon_power_plants (id, sources, source_id...
```

ClaraBuettner:
> The following error occurred while executing import-power-plants:
> 
> ```
> ERROR - (psycopg2.errors.UndefinedColumn) column "sources" of relation "egon_power_plants" does not exist
> LINE 1: INSERT INTO supply.egon_power_plants (id, sources, source_id...
> ```

I needed to change the former table structure, so the table is there and is not re-created but the columns have changed. Should I add a line to delete the old table first?

IlkaCu:
Might be the best to prevent other to run into this error.

IlkaCu (changes requested):
I deleted the table manually and re-run the power plant part - this worked fine. 
Please don't forget to update the changelog.

IlkaCu (approved):

PR #170, opened 2021-03-22
Drop tables before creation and data insertation

### LINKED ISSUE #166: Task 'zensus-inside-germany' fails
The task 'zensus-inside-germany' fails with the following error when the task already run before since the data is not deleted from the table. 

```
> psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "destatis_zensus_population_per_ha_inside_germany_pkey"
> DETAIL:  Key (gid)=(173) already exists.
```
I also noticed that the index of this table is still called 'gid' but we decided in #140 to rename it to 'id'.

gplssm:
Fixes #166

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -68,6 +68,8 @@ Changed
   `#148 <https://github.com/openego/eGon-data/issues/148>`_
 * Improve selection of zensus data in test mode
   `#151 <https://github.com/openego/eGon-data/issues/151>`_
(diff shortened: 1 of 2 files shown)


ClaraBuettner (approved):

ClaraBuettner:
With the new dev branch (and a new databse) the task fails again because the table you want to drop don't exists. Adding a `checkfirst=True` solves this. I could also fix this and reopen the branch or create a new one.

PR #173, opened 2021-03-22
Features/#167 adjust heat demand in unpopulated cells

### LINKED ISSUE #167: Deal with residential heat demand in cells without zensus population
There are some ha-cells with a residential heat demand but without a zensus population due to different datasets and methods. 
This results in issues, e.g. when creating heat demand time series based on the number of buildings from zensus. 
Until the methods from Peta can be integrated and directly use zensus population data, we will delete the residential heat demands in cells without zensus population and scale the demands in populated cells to meet the target values.

ClaraBuettner:
Closes #167


EvaWie (approved):
Thanks for the implementation of these adjustments!
I added a few changes in the formatting.

PR #176, opened 2021-03-23
Provide an example for updating to OEP-1.4.1 metadata standard

### LINKED ISSUE #139: Clarify metadata format
If I am not mistaking, we follow the [OEP Metadata version 1.4.0](https://github.com/OpenEnergyPlatform/oemetadata/blob/develop/metadata/v140/metadata_key_description.md).
Nevertheless, there seem to be some unclear points regarding the creation of metadata:
- How do we reference datasets, which have been created / loaded by other scripts into the local database?
- Which date is dataset creation date? How is that related to the version control of our datasets?

Please feel free to add further points.

gplssm:
This PR updates metadata for 

- VG250
- OpenStreetMap

Tasks we shouldn't forget:

- [x] Links for further details of Metadata needed to be added to docs
  - Example: https://github.com/OpenEnergyPlatform/oemetadata/blob/develop/metadata/latest/example.json
  - Description of metadata fields: https://github.com/OpenEnergyPlatform/oemetadata/blob/develop/metadata/latest/metadata_key_description.md
- [x] Move context to central place
- [x] Move linceses to central place
- [x] Adapt VG250 metadata according to decisions in https://github.com/openego/eGon-data/pull/176#issuecomment-808141771
- [x] Insert `resources` in OpenStreetMap metadata for each table. Use this script: https://github.com/openego/eGon-data/pull/176#issuecomment-865064295
- [x] Update CHANGELOG
- [x] Update docs

Fixes #139


gplssm:
I updated version of metadata to OEMetadata v1.4.1 on the example of VG250 tables. On the example of vg250_gem it looks like

```
'{
    "name": "boundaries.vg250_gem",
    "title": "BKG - Verwaltungsgebiete 1:250.000 - Gemeinden (GEM)",
    "description": "BKG - Verwaltungsgebiete 1:250.000 - Gemeinden (GEM)",
    "language": [
        "de-DE"
    ],
    "publicationDate": "2021-03-23",
    "context": {
        "homepage": "https://ego-n.org/",
        "documentation": "https://egon-data.readthedocs.io/en/latest/",
        "sourceCode": "https://github.com/openego/eGon-data",
        "contact": "https://ego-n.org/partners/",
        "grantNo": "03EI1002",
        "fundingAgency": "Bundesministerium f\u00fcr Wirtschaft und Energie",
        "fundingAgencyLogo": "https://www.innovation-beratung-foerderung.de/INNO/Redaktion/DE/Bilder/Titelbilder/titel_foerderlogo_bmwi.jpg?__blob=normal&v=3",
        "publisherLogo": "https://ego-n.org/images/eGon_logo_noborder_transbg.svg"
    },
    "spatial": {
        "location": null,
        "extent": "Germany",
        "resolution": "1:250000"
    },
    "temporal": {
        "referenceDate": "2020-01-01",
        "timeseries": {
            "start": null,
            "end": null,
            "resolution": null,
            "alignment": null,
            "aggregationType": null
        }
    },
    "sources": [
        {
            "title": "Verwaltungsgebiete 1:250 000 (Ebenen)",
            "description": "Der Datenbestand umfasst s\u00e4mtliche Verwaltungseinheiten der hierarchischen Verwaltungsebenen vom Staat bis zu den Gemeinden mit ihren Grenzen, statistischen Schl\u00fcsselzahlen, Namen der Verwaltungseinheit sowie die spezifische Bezeichnung der Verwaltungsebene des jeweiligen Landes.",
            "path": "https://daten.gdz.bkg.bund.de/produkte/vg/vg250_ebenen_0101/2020/vg250_01-01.geo84.shape.ebenen.zip",
            "licenses": [
                {
                    "name": "dl-by-de/2.0",
                    "title": "Datenlizenz Deutschland \u2013 Namensnennung \u2013 Version 2.0",
                    "path": "www.govdata.de/dl-de/by-2-0",
                    "instruction": "Jede Nutzung ist unter den Bedingungen dieser \u201eDatenlizenz Deutschland - Namensnennung - Version 2.0 zul\u00e4ssig.\nDie bereitgestellten Daten und Metadaten d\u00fcrfen f\u00fcr die kommerzielle und nicht kommerzielle Nutzung insbesondere:(1) vervielf\u00e4ltigt, ausgedruckt, pr\u00e4sentiert, ver\u00e4ndert, bearbeitet sowie an Dritte \u00fcbermittelt werden;\n (2) mit eigenen Daten und Daten Anderer zusammengef\u00fchrt und zu selbst\u00e4ndigen neuen Datens\u00e4tzen verbunden werden;\n (3) in interne und externe Gesch\u00e4ftsprozesse, Produkte und Anwendungen in \u00f6ffentlichen und nicht \u00f6ffentlichen elektronischen Netzwerken eingebunden werden.\nBei der Nutzung ist sicherzustellen, dass folgende Angaben als Quellenvermerk enthalten sind:\n(1) Bezeichnung des Bereitstellers nach dessen Ma\u00dfgabe,\n(2) der Vermerk Datenlizenz Deutschland \u2013 Namensnennung \u2013 Version 2.0 oder dl-de/by-2-0 mit Verweis auf den Lizenztext unter www.govdata.de/dl-de/by-2-0 sowie\n(3) einen Verweis auf den Datensatz (URI).Dies gilt nur soweit die datenhaltende Stelle die Angaben(1) bis (3) zum Quellenvermerk bereitstellt.\nVer\u00e4nderungen, Bearbeitungen, neue Gestaltungen oder sonstige Abwandlungen sind im Quellenvermerk mit dem Hinweis zu versehen, dass die Daten ge\u00e4ndert wurden.",
                    "attribution": "\u00a9 GeoBasis-DE / BKG (2021)"
                }
            ]
        }
    ],
    "licenses": [
        {
            "name": "dl-by-de/2.0",
            "title": "Datenlizenz Deutschland \u2013 Namensnennung \u2013 Version 2.0",
            "path": "www.govdata.de/dl-de/by-2-0",
            "instruction": "Jede Nutzung ist unter den Bedingungen dieser \u201eDatenlizenz Deutschland - Namensnennung - Version 2.0 zul\u00e4ssig.\nDie bereitgestellten Daten und Metadaten d\u00fcrfen f\u00fcr die kommerzielle und nicht kommerzielle Nutzung insbesondere:(1) vervielf\u00e4ltigt, ausgedruckt, pr\u00e4sentiert, ver\u00e4ndert, bearbeitet sowie an Dritte \u00fcbermittelt werden;\n (2) mit eigenen Daten und Daten Anderer zusammengef\u00fchrt und zu selbst\u00e4ndigen neuen Datens\u00e4tzen verbunden werden;\n (3) in interne und externe Gesch\u00e4ftsprozesse, Produkte und Anwendungen in \u00f6ffentlichen und nicht \u00f6ffentlichen elektronischen Netzwerken eingebunden werden.\nBei der Nutzung ist sicherzustellen, dass folgende Angaben als Quellenvermerk enthalten sind:\n(1) Bezeichnung des Bereitstellers nach dessen Ma\u00dfgabe,\n(2) der Vermerk Datenlizenz Deutschland \u2013 Namensnennung \u2013 Version 2.0 oder dl-de/by-2-0 mit Verweis auf den Lizenztext unter www.govdata.de/dl-de/by-2-0 sowie\n(3) einen Verweis auf den Datensatz (URI).Dies gilt nur soweit die datenhaltende Stelle die Angaben(1) bis (3) zum Quellenvermerk bereitstellt.\nVer\u00e4nderungen, Bearbeitungen, neue Gestaltungen oder sonstige Abwandlungen sind im Quellenvermerk mit dem Hinweis zu versehen, dass die Daten ge\u00e4ndert wurden.",
            "attribution": "\u00a9 GeoBasis-DE / BKG (2021)"
        }
    ],
    "contributors": [
        {
            "title": "Guido Ple\u00dfmann",
            "email": "http://github.com/gplssm",
            "date": "2020-12-04",
            "object": "",
            "comment": "Imported data"
        }
    ],
    "resources": [
        {
            "profile": "tabular-data-resource",
            "name": "boundaries.vg250_gem",
            "path": null,
            "format": "PostgreSQL",
            "encoding": "UTF-8",
            "schema": {
                "fields": [
                    {
                        "description": "Index",
                        "name": "gid",
                        "type": "integer",
                        "unit": null
                    },
                    {
                        "description": "Administrative level",
                        "name": "ade",
                        "type": "integer",
                        "unit": null
                    },
                    {
                        "description": "Geofactor",
                        "name": "gf",
                        "type": "integer",
                        "unit": null
                    },
                    {
                        "description": "Particular areas",
                        "name": "bsg",
                        "type": "integer",
                        "unit": null
                    },
                    {
                        "description": "Territorial code",
                        "name": "ars",
                        "type": "string",
                        "unit": null
                    },
                    {
                        "description": "Official Municipality Key",
                        "name": "ags",
                        "type": "string",
                        "unit": null
                    },
                    {
                        "description": "Seat of the administration (territorial code)",
                        "name": "sdv_ars",
                        "type": "string",
                        "unit": null
                    },
                    {
                        "description": "Geographical name",
                        "name": "gen",
                        "type": "string",
                        "unit": null
                    },
                    {
                        "description": "Designation of the administrative unit",
                        "name": "bez",
                        "type": "string",
                        "unit": null
                    },
                    {
                        "description": "Identifier",
                        "name": "ibz",
                        "type": "integer",
                        "unit": null
                    },
                    {
                        "description": "Note",
                        "name": "bem",
                        "type": "string",
                        "unit": null
                    },
                    {
                        "description": "Name generation",
                        "name": "nbd",
                        "type": "string",
                        "unit": null
                    },
                    {
                        "description": "Land (state)",
                        "name": "sn_l",
                        "type": "string",
                        "unit": null
                    },
                    {
                        "description": "Administrative district",
                        "name": "sn_r",
                        "type": "string",
                        "unit": null
                    },
                    {
                        "description": "District",
                        "name": "sn_k",
                        "type": "string",
                        "unit": null
                    },
                    {
                        "description": "Administrative association \u2013 front part",
                        "name": "sn_v1",
                        "type": "string",
                        "unit": null
                    },
                    {
                        "description": "Administrative association \u2013 rear part",
                        "name": "sn_v2",
                        "type": "string",
                        "unit": null
                    },
                    {
                        "description": "Municipality",
                        "name": "sn_g",
                        "type": "string",
                        "unit": null
                    },
                    {
                        "description": "Function of the 3rd key digit",
                        "name": "fk_s3",
                        "type": "string",
                        "unit": null
                    },
                    {
                        "description": "European statistics key",
                        "name": "nuts",
                        "type": "string",
                        "unit": null
                    },
                    {
                        "description": "Filled territorial code",
                        "name": "ars_0",
                        "type": "string",
                        "unit": null
                    },
                    {
                        "description": "Filled Official Municipality Key",
                        "name": "ags_0",
                        "type": "string",
                        "unit": null
                    },
                    {
                        "description": "Effectiveness",
                        "name": "wsk",
                        "type": "string",
                        "unit": null
                    },
                    {
                        "description": "DLM identifier",
                        "name": "debkg_id",
                        "type": "string",
                        "unit": null
                    },
                    {
                        "description": "Territorial code (deprecated column)",
                        "name": "rs",
                        "type": "string",
                        "unit": null
                    },
                    {
                        "description": "Seat of the administration (territorial code, deprecated column)",
                        "name": "sdv_rs",
                        "type": "string",
                        "unit": null
                    },
                    {
                        "description": "Filled territorial code (deprecated column)",
                        "name": "rs_0",
                        "type": "string",
                        "unit": null
                    },
                    {
                        "description": "Geometry of areas as WKB",
                        "name": "geometry",
                        "type": "Geometry(Polygon, srid=4326)",
                        "unit": null
                    }
                ],
                "primaryKey": [
                    "gid"
                ],
                "foreignKeys": null
            },
            "dialect": {
                "delimiter": null,
                "decimalSeparator": "."
            }
        }
    ],
    "metaMetadata": {
        "metadataVersion": "OEP-1.4.1",
        "metadataLicense": {
            "name": "CC0-1.0",
            "title": "Creative Commons Zero v1.0 Universal",
            "path": "https://creativecommons.org/publicdomain/zero/1.0/"
        }
    }
}'
```

Some fields couldn't properly be addressed. We need some decisions how we deal with these field in eGon-data

- id: not set, because neither URI nor DOI available yet. Should be added once published to OEP
- keywords: not used at the moment
- publicationDate was set to datetime.date.today(), we should decide if we do it differently or if we overwrite during publication on OEP 
- context should be provided centrally
- licenses should be written once as dictionary and then reused
- contributors: which changes are tracked here? In the VCS each of the changes is tracked, this could be doubled here, but maybe it's not worth. We should find a genreal aggreement in eGo^n
- resources > schema > fields > path not set because no publicly reachable URL available
- resources > schema > foreignKeys not set because no foreign keys are used in the example. Another example including FKs would be helpful

gplssm:
I have two more questions for OEP metadata experts. @christian-rli can I ask you again?

1. Does it make sense to fill all fields below "temporal" > "timeseries" or can we simply set "timeseries": null if it isn't a timeseries?
2. How to describe an empty field: null or "" (I chose null (None in python dicts respectively) for now)

christian-rli:
>     1. Does it make sense to fill all fields below "temporal" > "timeseries" or can we simply set "timeseries": null if it isn't a timeseries?

I thought about this just now and took a look at the parser and how different versions are shown on the website. Setting `"timeseries"` to `null` seems logical to me. It can also be parsed by OMI and the website will display it appropriately. Furthermore it is also possible to omit the key `"timeseries"` altogether.

>     2. How to describe an empty field: null or "" (I chose null (None in python dicts respectively) for now)

I'm not quite sure what a use-case might be that you are referring to. In which case would you need to reference a single empty field (or column?) in a data table within the metadata? Here are the metadata options I see: 

- `null` signifies that a specific metadata key (e.g. column description) does not apply for this specific data table or that the information wasn't supplied.
- If there is information available stating that in fact there is no such thing, an option could be to use "none". This could apply for example in a situation where there is no documentation available. A more descriptive value might also help, though.
- If the answer to one of the metadata fields is in fact empty, then I suppose an empty string "" could signify that... but I'd be interested in hearing the actual use case before recommending this option :)

gplssm:
Thanks for your answer!

> > ```
> > 1. Does it make sense to fill all fields below "temporal" > "timeseries" or can we simply set "timeseries": null if it isn't a timeseries?
> > ```
> 
> I thought about this just now and took a look at the parser and how different versions are shown on the website. Setting `"timeseries"` to `null` seems logical to me. It can also be parsed by OMI and the website will display it appropriately. Furthermore it is also possible to omit the key `"timeseries"` altogether.

Thanks for checking this! It seems to me that omitting keys is the most plausible approach.

> > ```
> > 2. How to describe an empty field: null or "" (I chose null (None in python dicts respectively) for now)
> > ```
> 
> I'm not quite sure what a use-case might be that you are referring to. In which case would you need to reference a single empty field (or column?) in a data table within the metadata? Here are the metadata options I see:
> 
>     * `null` signifies that a specific metadata key (e.g. column description) does not apply for this specific data table or that the information wasn't supplied.
> 
>     * If there is information available stating that in fact there is no such thing, an option could be to use "none". This could apply for example in a situation where there is no documentation available. A more descriptive value might also help, though.
> 
>     * If the answer to one of the metadata fields is in fact empty, then I suppose an empty string "" could signify that... but I'd be interested in hearing the actual use case before recommending this option :)

Ok, thanks for this detailed explanation. Indeed, it might make sense to use a field for explaining why there is no information. For example, in case of a missing documentation.

Takeaway for use: omit the key or set the value to `None` if the key isn't applicable (for example `timeseries` in case of non-timeseries data). Both should results in the same data being displayed on the OEP.
Using an empty string `""` doesn't make too much sense. 

Btw, you mentioned OMI. Does it maybe make sense to use OMI in eGon-data for testing metadata string parsing?

christian-rli:
> Takeaway for use: omit the key or set the value to `None` if the key isn't applicable (for example `timeseries` in case of non-timeseries data). Both should results in the same data being displayed on the OEP.

Just for the sake of clarity `null` does not equal `"none"`. `null` means there is nothing here because it doesn't apply ("this isn't a timeseries, so providing any information here is out of place") or this information is left blank ("this might be a timeseries, but nobody specifically said so"). 
My initial interpretation of `"none"` would be that one might reasonably expect information here, but in fact we're told that such information does not exist (as in the documentation example). The second interpretation that you might have understood from my comment (unless it was simply a typo) is that `"none"` explicitly states that this metadata field does not apply and that it's not just a field that was left empty.

For now, I would like to ask you not to go with the second interpretation, but to omit the field or enter `null` in cases where it doesn't apply. I do see the case that one might want to state that this field explicitly does not apply and wasn't just left empty. However I would rate the importance of this use case lower than the extra mental effort or manual labor for explicit labelling (until someone convinces me otherwise)... I realize that this that this system of emptiness, none, null, and potentially "" can be confusing and I'd be happy to switch to a system that is more intuitive. I'd welcome any comments and suggestions regarding this matter! Also, in case my comments were somehow unclear or difficult to follow.

[OMI](https://github.com/OpenEnergyPlatform/omi) is the tool that is used by the OEP to parse the metadata (this includes btw a converter from version 1.3 to 1.4.). Before uploading to the OEP I usually parse my strings with OMI to see if they are going to be accepted beforehand. So if you are going to have a lot of strings it would indeed make sense to check them automatically. In my experience OMI isn't the most userfriendly program, but the basic command that I use for parsing my strings is:

`omi translate -f oep-v1.4 my_string.json`

I'd be very happy to receive bug reports, suggestions for better error messages and more general question marks, because it would help a lot with improving the tool.

gplssm:
> > Takeaway for use: omit the key or set the value to `None` if the key isn't applicable (for example `timeseries` in case of non-timeseries data). Both should results in the same data being displayed on the OEP.
> 
> Just for the sake of clarity `null` does not equal `"none"`. `null` means there is nothing here because it doesn't apply ("this isn't a timeseries, so providing any information here is out of place") or this information is left blank ("this might be a timeseries, but nobody specifically said so").
> My initial interpretation of `"none"` would be that one might reasonably expect information here, but in fact we're told that such information does not exist (as in the documentation example). The second interpretation that you might have understood from my comment (unless it was simply a typo) is that `"none"` explicitly states that this metadata field does not apply and that it's not just a field that was left empty.
> 
> For now, I would like to ask you not to go with the second interpretation, but to omit the field or enter `null` in cases where it doesn't apply. I do see the case that one might want to state that this field explicitly does not apply and wasn't just left empty. However I would rate the importance of this use case lower than the extra mental effort or manual labor for explicit labelling (until someone convinces me otherwise)... I realize that this that this system of emptiness, none, null, and potentially "" can be confusing and I'd be happy to switch to a system that is more intuitive. I'd welcome any comments and suggestions regarding this matter! Also, in case my comments were somehow unclear or difficult to follow.

Thanks for this detailed explanation again :+1: 
I think I  can understand the difference and I agree: using `null` or omitting a key is favorable. This works well for us and, as far as I can think of future cases, we won't need the `"none"` option.

> [OMI](https://github.com/OpenEnergyPlatform/omi) is the tool that is used by the OEP to parse the metadata (this includes btw a converter from version 1.3 to 1.4.). Before uploading to the OEP I usually parse my strings with OMI to see if they are going to be accepted beforehand. So if you are going to have a lot of strings it would indeed make sense to check them automatically. In my experience OMI isn't the most userfriendly program, but the basic command that I use for parsing my strings is:
> 
> `omi translate -f oep-v1.4 my_string.json`
>
> I'd be very happy to receive bug reports, suggestions for better error messages and more general question marks, because it would help a lot with improving the tool.

Thanks also for this explanation! I agreed with some members of the consortium that it would make sense from our perspective to include this check with OMI.  
We won't do this the next days, but maybe in April. We'll keep you updated.

gplssm:
@christian-rli one more question :)

We've talked about the field `keywords` and asked ourselves if these are related to tags shown on the OEP. Is there any relation between keywords and tags? Are the tags generated from keywords or is this something completely different?

christian-rli:
> @christian-rli one more question :)
> 
> We've talked about the field `keywords` and asked ourselves if these are related to tags shown on the OEP. Is there any relation between keywords and tags? Are the tags generated from keywords or is this something completely different?

The idea is that there _should_ be a connection. That would be an easy way to work with tags. However for historical reasons this isn't implemented (yet). Tags were integrated into the OEP before the metadata-string. They are database entries with ids and all that, while keywords technically are part of a comment on the table. An integration now would/will be a bit tricky. Tags could arise multiple times and a connection would need to be built between comments and the tags table - in both directions. So while there is no conceptual difference between the two, they are technically separate and need to be updated in different ways. The only way I know how to update tags is through the OEP interface. Keywords can be updated via api and interface, but only tags are shown next to the table and can be used to filter in the interface.

I was positive that there would be an issue for the connection of the two, but i couldn't find it [so I opened one](https://github.com/OpenEnergyPlatform/oeplatform/issues/740).

gplssm:
Thanks for the detailed answer again!

I can confirm that a connection between keywords and tags would be useful. 
However, it's not a super important feature we have to use. We'll leave the keywords empty for now.

gplssm:
Decisions for fields that were listed in https://github.com/openego/eGon-data/pull/176#issuecomment-805049720 from telco on 25th of March:

- id: must be added during publication on OEP
- keywords: will not be used
- publicationDate: must be updated during publication on OEP
- contributors: we list all contributors who contributed creating the described data. The fields `date`, `obj` and `comment` will be left blank. `email` is optional
- resources > schema > fields > path: must be added during publication on OEP
- resources > schema > foreignKeys: Another example including FKs would be helpful

christian-rli:
> resources > schema > foreignKeys: Another example including FKs would be helpful

@jh-RLI has provided me with an [example json file](https://github.com/OpenEnergyPlatform/data-preprocessing/blob/44cdc13276cf942fc91af2153b53aade3c492e66/data-review/open_MODEX_base_scenario_frame_datapackage.json) that covers several resources in one metadata file and connects those with a foreign key.

Table creation is done with [oem2orm](https://github.com/OpenEnergyPlatform/oem2orm). If you need more details, ask @jh-RLI who has more experience with this example and who also developed oem2orm.

gplssm:
Script used to extract metadata from database table

```
from sqlalchemy import BigInteger, BIGINT

from egon.data.processing.zensus_vg250.zensus_population_inside_germany import Vg250Sta



fields = []


sqlalchemy_type_map = {
    "BIGINT": "integer",
    "VARCHAR": "string"
}

for col in Vg250Sta.__table__.columns:
    print("\"" + str(col).split(".")[1] + ":\" ")

    field = {
        "name": col.name,
        "description": "",
        "type": sqlalchemy_type_map.get(str(col.type), col.type),
        # "type": sqlalchemy_type_map[str(col.type)],
        "unit": None
    }

    fields.append(field)

import pprint

pprint.pprint(fields)
```

nesnoj:
@gnn this is the mentioned workaround for the reflection of geometry columns:
https://github.com/openego/eGon-data/blob/7229d55e8af7bbfdb7fb12fa8592aa3a9c2f06cd/src/egon/data/metadata.py#L225-L240

nesnoj:
@gnn may I request your review?

CONTRIBUTING.rst
```diff
@@ -275,6 +275,69 @@ be saved locally, please use `CWD` to store the data. This is achieved by using
...
+
+Helpers
+^^^^^^^
+
+There are some **licence templates** provided in :py:mod:`egon.data.metadata`
+you can make use of for fields 11.4 and 12 of the
+`Open Energy Metadata Description`_. Also, there's a template for the
+**metaMetadata** (field 16).
+
+[Some content about **sources** will be inserted here]
```
ClaraBuettner, line 328: I'm wondering if we could already write something about our today's meeting of the manual sources-list here. What do you think?

src/egon/data/metadata.py
```diff
@@ -0,0 +1,262 @@
...
+    Returns
+    -------
+    list of dict
+        Resource fields
+    """
+
+    # TODO: Workaround as SQLA does not recognize Geometry columns. This only
+    #  covers the dtype of a specific column 'geom', how can this be done in a
+    #  generic way?
+    ischema_names['geom'] = Geometry
```
ClaraBuettner, line 250: As you already explained in the meeting last week this might be a bit problematic for SQL-defined tables with another geom column name. Do you want to find a solution (e.g. giving the geom-column names as an input-parameter to the function) before this is merged or prefer to solve it when we run into problems?

src/egon/data/datasets/osm/__init__.py
```diff
@@ -114,7 +120,9 @@ def to_postgres(num_processes=1, cache_size=4096):
 
 
 def add_metadata():
-    """Writes metadata JSON string into table comment."""
+    """Writes metadata JSON string into table comment.
```
ClaraBuettner, line 123: In my case the osm-Dataset was skipped because the version already existed. I started a clean run in an empty database now to make sure it is executed, but I thought that the Dataset-version should be always updated when the code changed. Or am I mistaken?

CONTRIBUTING.rst
nesnoj: Good point, I'll add this to the docs.

src/egon/data/metadata.py
nesnoj: I'd implement the geom-columns-as-input-params as minimum solution.
I haven't put more efforts into a more generic approach, what about u @gnn?

src/egon/data/datasets/osm/__init__.py
nesnoj: \*cough\* yes, I forgot that, thanks.

ClaraBuettner:
I tested this branch successfully in testmode in an empty database and new environment.
In addition I tested the omi-tool as explained in the docs, it was also working as expected.

CONTRIBUTING.rst
nesnoj: Done in 802f86b0bf2cad41559a41e84c499755bc161872

ClaraBuettner: Yes, thank you!

src/egon/data/metadata.py
nesnoj: > I'd implement the geom-columns-as-input-params as minimum solution.

Done in ec7d5eefdd2bc4eae21bb68135da23d527aa3bfc

ClaraBuettner (approved):

nesnoj:
Thanks a lot for your review @ClaraBuettner !

PR #159, opened 2021-03-15
Allow configuring the database name

gnn:
If the configured database doesn't exist, it will be created. This enables seamless usage of multiple databases and should fix #112.

PR #180, opened 2021-03-24
Features/#165 centralize map zensus to boundaries

### LINKED ISSUE #165: Map zensus cells to bounaries only once
The zensus cells are mapped to boundaries.vg250_gem ([inside a SQL-query](https://github.com/openego/eGon-data/blob/fc1adc0437c0df8197f2fede59c0829686962684/src/egon/data/processing/zensus_vg250/zensus_population_inside_germany.py#L240-L244)) and to boundaries.vg250_krs ([here](https://github.com/openego/eGon-data/blob/fc1adc0437c0df8197f2fede59c0829686962684/src/egon/data/processing/zensus/__init__.py#L73)). 

When I wrote the second one, I did not knew that the first one will be added later on. As @IlkaCu said, I could rewrite it and use only the zensus cells inside Germany. But we would still have two spatial joins combining zensus cells and vg250 which need about 12 or 6 minutes only for Schleswig-Holstein. 
I would prefer to do this only once and create a table which maps all (populated) zensus cells to the boundaries.vg250_gem and use this in all function which need it.
@gplssm @IlkaCu: What do you think?

ClaraBuettner:
This branch adds a table which maps zensus data to vg250 municipalities and nuts3 regions. 
The table is used by the society prognosis and the population_in_municipalities. 
@gplssm Since it wasn't that easy for me to adjust your SQLalchemy code, I re-wrote it with (geo)pandas because it was much easier for me and also needed less lines of code. Could you please take a look and tell me if you are fine with this? 

As I also posted in #165 I decided to map all zensus cells with a population to municipalities. 

Fixes #165

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -85,6 +85,8 @@ Changed
   `#166 <https://github.com/openego/eGon-data/issues/166>`_
 * Adjust residential heat demand in unpopulated zenus cells
   `#167 <https://github.com/openego/eGon-data/issues/167>`_
+* Introduce mapping between VG250 municipalities and census cells
+  `#165 <https://github.com/openego/eGon-data/issues/165>`_
 
 Bug fixes
 ---------
--- a/src/egon/data/airflow/dags/pipeline.py
+++ b/src/egon/data/airflow/dags/pipeline.py
@@ -133,6 +133,11 @@ with airflow.DAG(
     population_import >> zensus_misc_import
 
     # Combine Zensus and VG250 data
+    map_zensus_vg250 = PythonOperator(
+        task_id="map_zensus_vg250",
+        python_callable=zensus_vg250.map_zensus_vg250,
+    )
+
     zensus_inside_ger = PythonOperator(
         task_id="zensus-inside-germany",
         python_callable=zensus_vg250.inside_germany,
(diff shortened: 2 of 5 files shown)


ClaraBuettner:
All 'Tests, code style & coverage' failed. But looking into the log, I don't think that this is caused by my changes. It looks more like this happens because the file `egon-data.pid-...yml ` was moved to the current working directory in #159.

gplssm (approved):
Looks good!

I tested locally and had a quick look on vg250_gem_population that's now created via the mapping table. Seems plausible!

Thanks for making these changes!

Regarding the implementation in geopandas rather than in postgis. That's totally fine! Geopandas made significant performance improvements recently (2019 I guess) by using pygeos directly. Now, geopandas should be as fast as postgis or even faster.

Since the issue about tests is already described in https://github.com/openego/eGon-data/issues/183, that's fine and not our problem here.

Regarding the comment on the implementation in https://github.com/openego/eGon-data/issues/165#issuecomment-806011537: I'm generally fine with that. I might lead to different number of inhabitants depending on which table is used. If you ask vg250_gem_population you probably get a higher value compared to create the sum over all census cells that are located with the centroid inside Germany (namely destatis_zensus_population_per_ha_inside_germany). But this is actually a matter of how data is used subsequently.

I'll a note to the CHANGELOG an gonna merge.

PR #182, opened 2021-03-25
Delete zensus-inside-germany tables only if they exist

ClaraBuettner:


EvaWie (approved):
The workflow does not fail anymore.

PR #184, opened 2021-03-25
Features/#181 electrical demands to zensus cells

### LINKED ISSUE #181: Distribute electrical demands of households to zensus cells
The electrical demands for households of demandregio (nuts3 level) are distributed to the zensus cells. This is done linear to the share of population in each zensus cell per nuts3 region.
By implementing this, I will also divide the 'import-demandregio' task into different tasks to allow parallel application.

ClaraBuettner:
Closes #181 
This branch adds a function which distributes the electrical demand of households to zensus cells. 
I also divided the 'import-demandregio' task into separate tasks for importing society data and household and cts/industry demands.


ClaraBuettner:
I renamed the variable and dateset configuration key in 0537420 and 9b5a47c

IlkaCu (commented):
I am wondering if we can find a more comprehensible name for table demand.egon_demandregio_household_electricity which differs a bit more from demand.egon_demandregio_hh

ClaraBuettner:
> I am wondering if we can find a more comprehensible name for table demand.egon_demandregio_household_electricity which differs a bit more from demand.egon_demandregio_hh

I also thought about this. But demand.egon_demandregio_household_electricity was already such a long name that I didn't wanted to add zensus to it. 
But maybe we can think of writing cts and household demands into this table (as also done for the heat demands) and call it demand.egon_demandregio_zensus_electricity?

ClaraBuettner:
I renamed and re-structured the table as we discussed in f9550a7
In addition, I noticed that the population_prognosis is ~10 times faster when only the populated cells are imported. So I decided to insert only them in 2efb399

IlkaCu (approved):

PR #187, opened 2021-03-29
Fix automatic checks

### LINKED ISSUE #183: GitHub actions fail
All GitHub actions of PR that include the current dev fail with the following error: 

```
> 
>   ==================================== ERRORS ====================================
>   _ ERROR collecting .tox/py37-nocov-linux/lib/python3.7/site-packages/egon/data/airflow/dags/pipeline.py _
>   .tox/py37-nocov-linux/lib/python3.7/site-packages/egon/data/airflow/dags/pipeline.py:30: in <module>
>       airflow_db_connection()
>   .tox/py37-nocov-linux/lib/python3.7/site-packages/egon/data/db.py:108: in airflow_db_connection
>       cred = credentials()
>   .tox/py37-nocov-linux/lib/python3.7/site-packages/egon/data/db.py:28: in credentials
>       custom = config.paths(pid="*")[0]
>   E   IndexError: list index out of range
```

gnn:
Fixes #183 by removing a module global function call and putting the corresponding logic into `egon-data`'s startup function, using central configuration `settings` retrieval functionality and adjusting tests to previous changes.

### DIFF
--- a/src/egon/data/cli.py
+++ b/src/egon/data/cli.py
@@ -259,8 +259,8 @@ def egon_data(context, **kwargs):
     #       import can only be done inside this function, which is generally
     #       frowned upon, instead of at the module level. Maybe there's a
     #       better way to encapsulate this?
-    from airflow.configuration import conf as airflow_cfg
-    from airflow.models import Connection
+    from airflow.configuration import conf as airflow_cfg  # noqa
+    from airflow.models import Connection  # noqa
 
     options = options["egon-data"]
     render(


gnn:
I merged part of this PR into dev because @gplssm needed the fix. So for now, issue #183 is closed, but this PR remains open until @ClaraBuettner and @EvaWie confirm that #183 is indeed fixed.

ClaraBuettner (approved):
Yes, as I commented in #183 , I tested it and everything worked fine. 
But there is no note in the changelog about this.

gnn:
> Yes, as I commented in #183 , I tested it and everything worked fine.

Cool. So that leaves the question of whether you, @EvaWie, are able to test your code again after the fix.

> But there is no note in the changelog about this.

Maybe it's time for a release? Bug fixes usually only get noted in the changelog when they fix bugs appearing in a previous release. Bugs which appear during development of a particular release and are also fixed during the same development cycle usually don't get a note in the changelog because in theory, regular users never see the bug. In general, the changelog lists changes from one release to another.
So, since we've been developing the current version for a while now, maybe it's time to start adopting a biweekly or monthly release cycle to get a periodic feature freeze and bug status?

ClaraBuettner:
> Maybe it's time for a release? Bug fixes usually only get noted in the changelog when they fix bugs appearing in a previous release. Bugs which appear during development of a particular release and are also fixed during the same development cycle usually don't get a note in the changelog because in theory, regular users never see the bug. In general, the changelog lists changes from one release to another.
> So, since we've been developing the current version for a while now, maybe it's time to start adopting a biweekly or monthly release cycle to get a periodic feature freeze and bug status?

If we want to release on pypi and not only on github, I need to adjust one thing (the desaggregator installation), but I will do this soon. Besides that, I think there is nothing against a release.

gplssm:
> > Maybe it's time for a release? Bug fixes usually only get noted in the changelog when they fix bugs appearing in a previous release. Bugs which appear during development of a particular release and are also fixed during the same development cycle usually don't get a note in the changelog because in theory, regular users never see the bug. In general, the changelog lists changes from one release to another.
> > So, since we've been developing the current version for a while now, maybe it's time to start adopting a biweekly or monthly release cycle to get a periodic feature freeze and bug status?
> 
> If we want to release on pypi and not only on github, I need to adjust one thing (the desaggregator installation), but I will do this soon. Besides that, I think there is nothing against a release.

_Releasing_ is a bit ambiguous. I can refer to create a release on GitHub (via git tags) or to publishing a package on pypi. Maybe it helps to distinguish these two.

From my point of view, it is helpful to create a GitHub release (aka a version tag of eGon-data) in order to get a better overview on fixed bugs and changes in general. We could do this very soon, if someone is interested in doing it ;-)

Publishing the package on pypi makes sense, once we can provide useful data for at least one model that uses eGon-data generated data (eTraGo, ding0, ..). 

This means, no hurry @ClaraBuettner with doing the changes. Unless you're done with the eTraGo input data.

EvaWie:
> Cool. So that leaves the question of whether you, @EvaWie, are able to test your code again after the fix.

Thanks for solving this issue! I  can now test my code parts, also when airflow is not running.

PR #195, opened 2021-04-01
Features/#15 osmtgmod to pypsa

ulfmueller:
integrates the old data processing step "ego_dp_powerflow_osmtgmod_to_pypsa.sql" into the work flow.

The version and the scenario name implementation is very beta and should be tackled at some point.


src/egon/data/processing/osmtgmod/osmtgmod_to_pypsa.sql
```diff
@@ -0,0 +1,172 @@
...
+	(SELECT bus0 FROM grid.egon_pf_hv_line WHERE scn_name='Status Quo')
+AND bus_id NOT IN 
+	(SELECT bus1 FROM grid.egon_pf_hv_line WHERE scn_name='Status Quo')
+AND bus_id NOT IN 
+	(SELECT bus0 FROM grid.egon_pf_hv_transformer WHERE scn_name='Status Quo')
+AND bus_id NOT IN 
+	(SELECT bus1 FROM grid.egon_pf_hv_transformer WHERE scn_name='Status Quo'); 
+	
+
+/*
```
IlkaCu, line 108: Is there any need or benefit to keep the code which is commented out?

ulfmueller: u r right. I skipped that part in the newest version.

ClaraBuettner:
I tested this branch and everything worked fine. The result looks good (the topology is a little different compared to our former grid model, but that's another task...)
Could you add a note to the changelog?

ulfmueller:
I think the task is to minor for a changelog entry, I would say....

PR #197, opened 2021-04-06
Fix zensus-inside-germany with --dataset-boundary=Everything

### LINKED ISSUE #196: The `"zensus-inside-germany"` task fails with `--dataset-boundary=Everything`
There are other failures, but this one is harder to track down because the log file is to big to be viewed easily. Here's the error, before it gets too long:

```
[2021-04-01 02:14:15,939] {taskinstance.py:1150} ERROR - (psycopg2.errors.InternalError_) invalid memory alloc request size 1073741824

[SQL: INSERT INTO society.destatis_zensus_population_per_ha_inside_germany (gid, grid_id, population, geom_point, geom) SELECT society.destatis_zensus_population_per_ha.id AS society_destatis_zensus_population_per_ha_id, society.destatis_zensus_population_per_ha.grid_id AS society_destatis_zensus_population_per_ha_grid_id, society.destatis_zensus_population_per_ha.population AS society_destatis_zensus_population_per_ha_population, society.destatis_zensus_population_per_ha.geom_point AS society_destatis_zensus_population_per_ha_geom_point, society.destatis_zensus_population_per_ha.geom AS society_destatis_zensus_population_per_ha_geom
FROM society.destatis_zensus_population_per_ha
WHERE society.destatis_zensus_population_per_ha.population > %(population_1)s AND society.destatis_zensus_population_per_ha.id IN (...
```

The array after the `IN (` is way too long to display. I think the problem lies in the fact that the [query][0] retrieving the `cells_in_germany` selects too much data. This is what's placed in the array after the `IN (`. The log file is around 2.5G big. So the subsequent `query` using `.filter(DestatisZensusPopulationPerHa.id.in_(cells_in_germany))` is issuing a 2.5G large select statement to the database. It's a good idea to avoid such data round trips between the database and the Python process whenever possible and filter with a subquery or via a join instead so the data remains inside the database process for as long as possible.

[0]: https://github.com/openego/eGon-data/blob/de2f63d909ad13ddaaf0bfd98284786d6664e267/src/egon/data/processing/zensus_vg250/zensus_population_inside_germany.py#L237-L240

ClaraBuettner:
I think this fixes #196 but I was not able to test it with` --dataset-boundary=Everything` . @gnn Could you try to run it?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -96,6 +96,8 @@ Changed
   `#166 <https://github.com/openego/eGon-data/issues/166>`_
 * Add gdal to pre-requisites
   `#185 <https://github.com/openego/eGon-data/issues/185>`_
+* Update task zensus-inside-germany
+  `#196 <https://github.com/openego/eGon-data/issues/196>`_
 
 Bug fixes
 ---------
--- a/src/egon/data/processing/zensus_vg250/zensus_population_inside_germany.py
+++ b/src/egon/data/processing/zensus_vg250/zensus_population_inside_germany.py
@@ -233,13 +233,11 @@ def inside_germany():
         bind=engine_local_db, checkfirst=True
     )
 
-    # Selects zensus cells in German boundaries from vg250
-    cells_in_germany = db.select_dataframe(
-        """SELECT zensus_population_id
-        FROM boundaries.egon_map_zensus_vg250"""
-    ).zensus_population_id.values.tolist()
-
     with db.session_scope() as s:
+
+        # Query zensus cells in German boundaries from vg250
+        cells_in_germany = s.query(MapZensusVg250.zensus_population_id)
+
         # Query relevant data from zensus population table
         q = (
             s.query(


gnn:
> @gnn Could you try to run it?

I used this occasion to create an account on the server and insert your SSH key. Check your email. So you should be able to test it yourself. But I'll be trying it out, too. At first glance it looks OK for me though.

gnn:
Looks like the fix worked. The task completed without an error.

gnn:
Btw. you can use `ssh -L 8080:SERVER:8080 USER@SERVER` to forward your local port 8080 to the one on the server. That way you can start `egon-data serve` on the server and still use your local browser to control the pipeline like you would've done if you started `egon-data serve` on your local machine.

ClaraBuettner:
Thank's for testing and the hint with the forwarded port! 
I tested the workflow with `--dataset-boundary=Everything` and `zensus-inside-germany` doesn't fail anymore. The results look fine, so I will merge this into dev.

PR #205, opened 2021-04-09
Fixes/#201 change demandregio nuts names

### LINKED ISSUE #201: Demandregio uses older version of nuts regions
eGon-data imports nuts regions in version 2016. demandRegio uses the the regions from 2013, they added the one nuts3-region which is new in version 2016, to be compatible. But still there are problems with two nuts3-regions (namely DEB16 and DEB19) which slightly changed in version 2016 and got new names (DEB1C and DEB1D). 
I checked it in qGis, the new and the old boundaries look very similar so I think we can just rename them. 
The errors are in population_prognosis. I could fix them there, but I guess we might get problems with the different names in other tasks as well. I'm wondering if I should rename the nuts3-regions of demandregio directly in the import. 
@IlkaCu what do you think?

ClaraBuettner:
Fixes #201 
I changed the names of two demandregio nuts3 regions which were renamed in nuts3 version 2016.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -105,3 +105,5 @@ Bug fixes
   `#157 <https://github.com/openego/eGon-data/issues/157>`_
 * Substation sequence
   `#171 <https://github.com/openego/eGon-data/issues/171>`_
+* Adjust names of demandregios nuts3 regions according to nuts version 2016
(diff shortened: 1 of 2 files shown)


IlkaCu:
Please don't forget to update the changelog 😉

IlkaCu (approved):

PR #208, opened 2021-04-09
Place files created during the workflow below the current working directory

### LINKED ISSUE #203: Create new files inside the working directory
Currently most of the newly created files are created inside the installation directory of the package. This is not a good idea, as this directory might not even be writable by regular users. In order to fix this, new files should be created in the working directory or a subdirectory of the working directory.

### LINKED ISSUE #341: Store files in working dir
In some tasks, files are stored in the a task's directory instead of the working dir which induces subsequent problems and makes the pipeline more error-prone.
This came up in https://github.com/openego/eGon-data/pull/338#discussion_r673135401 , check out for more background info.

- check if your dataset is affected
- modify your task, use a proper way to get current working dir (see link above how to do that)
- adjust downstream tasks accordingly (or notify responsible person)

This should be approached as soon as all tasks will have been migrated to datasets.

@gnn Please amend if I've missed something here..

gnn:
Fixes #341.

### DIFF
--- a/src/egon/data/importing/openstreetmap/__init__.py
+++ b/src/egon/data/importing/openstreetmap/__init__.py
@@ -9,9 +9,9 @@ If you have to import code from a module below this one because the code
 isn't exported from this module, please file a bug, so we can fix this.
 """
 
+from pathlib import Path
 from urllib.request import urlretrieve
 import json
-import os
 import time
 
 from egon.data import db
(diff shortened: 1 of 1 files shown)


nesnoj:
Some parts may be outdated and more recent datasets are not covered. But it's a good basis to start from to solve #341..

nailend:
This PR might conflict with changes I did in #341. As @gnn was using a different code base, I copied the differences manually.

PR #213, opened 2021-04-13
Fixes/#204 heat demand import everything

### LINKED ISSUE #204: Task import-heat-demand fails with --dataset-boundary=Everything
The heat demand import fails with the following log, the log file is ~4GB so I wasn't even able to open it completely, but this screenshot shows the most important lines (just the huge list of values is missing). 
log_import-heat-demand

It looks the these lines need too much memory: https://github.com/openego/eGon-data/blob/64b79217c5a203aa5b9278d6be8c3586f5c4005f/src/egon/data/importing/heat_demand_data/__init__.py#L478-L489

I think we can't really change something in these lines to reduce the memory because it is already directly inserted to the database. 
But regarding htop, the maximal RAM used is ~20GB and the server has much more.
@gnn Is there a limit for airflow or the docker container and could we try to adjust this?

ClaraBuettner:
Fixes #204 
Adding the `-t auto` as suggested in https://github.com/openego/eGon-data/issues/204#issuecomment-817943065 didn't help. So I  import the raster data for each sector and scenario (so each raster-file) separately which solves the problem. 
@gnn: Is it okay to do it like this? 
@EvaWie: Could you please test this (for Schleswig-Holstein)?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -113,3 +113,5 @@ Bug fixes
   `#201 <https://github.com/openego/eGon-data/issues/201>`_
 * Delete zensus buildings, apartments and households in unpopulated cells
   `#202 <https://github.com/openego/eGon-data/issues/202>`_
+* Import heat demand raster files successively to fix import for dataset==Everything
+  `#204 <https://github.com/openego/eGon-data/issues/204>`_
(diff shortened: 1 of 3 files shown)


EvaWie:
Thanks for this nice idea, solving the problem.
I run the workflow for Schleswig-Holstein and it worked.

PR #224, opened 2021-04-21
change to a more actual osm dataset

### LINKED ISSUE #223: Change osm data import to the 2021-01-01 datasets
change in the datasets.yaml the osm data downloads.
For Germany: http://download.geofabrik.de/europe/germany-210101.osm.pbf
For test mode /Schleswig-Holstein: http://download.geofabrik.de/europe/germany/schleswig-holstein-210101.osm.pbf

ulfmueller:
closes #223

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -35,6 +35,8 @@ Added
 
 * OSM data import as done in open_ego
   `#1 <https://github.com/openego/eGon-data/issues/1>`_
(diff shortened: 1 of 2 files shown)


gplssm (approved):
Thanks for updating!

I've tested with Schleswig-Holstein on my computer and it works well. I assume it also works for Germany.

I didn't checked if the updated data lands in the database, but I'm quite sure about that this works ;-)

PR #225, opened 2021-04-21
Integrate scenario table

ClaraBuettner:
This branch adds a table called "scenario.egon_sceanrio_parameters" which includes parameters for each sector. These parameters can be used in other tasks (as done as an example for demandregio-import). 
All parameters written into the dictionaries inside parameters.py are added to the table.


ClaraBuettner:
@EvaWie could you test this branch and check if the heat parameters are added correctly?

EvaWie (approved):
I had a look at the changes and run the test case. Everything worked fine.
Thanks a lot for this nice implementation!

PR #229, opened 2021-04-27
Fix failing pgrouting installation by updating sources before

### LINKED ISSUE #115: include pgrouting extension
For #4 it is necessary to have the extension pgrouting. Unfortunately, so far I was not able to get this extension working. osmTGmod throws an error stating that the pgrouting.control file is missing in the extensions folder of postgresql (/usr/share/postgresql/12/extension).

I tried numerous things. Withing egon-data I tried out 31af5bdbf9b3b7e3d58a26f31add64b425793650

Furthermore I tried this: https://docs.pgrouting.org/3.0/en/pgRouting-installation.html
It is failing because boost was not available when executing the cmake command. Installing boost did not work either.

I also found this: https://github.com/pgRouting/docker-pgrouting
which so far did not lead to anything really. To change our docker to this image seems to be most promising to me though right now.

gplssm:
After removing the container and the related volume, hence, recreating everything from scratch, I ran into the problem that was described in https://github.com/openego/eGon-data/issues/115#issuecomment-810983858 again. In order to avoid further trouble here, I propose this one line fix. 
I would call it harmless. So, if you consider to approve without testing locally, I think it is ok ;-)

Fixes #115 (again).

### DIFF
--- a/src/egon/data/airflow/Dockerfile.postgis
+++ b/src/egon/data/airflow/Dockerfile.postgis
@@ -2,6 +2,7 @@ FROM postgres:12
 
 RUN apt-get update
 RUN apt-get install -y postgresql-12-postgis-3
+RUN apt-get update
 RUN apt-get install -y postgresql-12-pgrouting
 
 


gplssm:
Hey @ulfmueller can I ask you for a quick review here? This change should affect to many people. But on the RLI side, we would benefit regarding workflow execution on our server

gplssm:
As this change should harmless and it helps me to avoid stashing back and forth, I gonna merge it now. Regardless of any review.

PR #230, opened 2021-04-28
Store database data in the current working directory

### LINKED ISSUE #228: Allow for custom docker image location
It'd be great to allow for a custom location where the docker image is to be stored, [docker-compose](https://docs.docker.com/compose/reference/) allows to do so.
Could we add a CLI option for this, what do you think @gnn ?

gnn:
Closes #228. I think the underlying issue of why the feature was requested should be eliminated by this change. If that's not the case, the PR can be extended to fully fix the issue.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -76,6 +76,12 @@ Added
   `#117 <https://github.com/openego/eGon-data/issues/117>`_
 * Integrate scenario table with parameters for each sector 
   `#177 <https://github.com/openego/eGon-data/issues/177>`_
+* The volume of the docker container for the PostgreSQL database
+  is saved in the project directory under `docker/database-data`.
+  The current user (`$USER`) is owner of the volume.
+  Containers created prior to this change will fail when using the 
+  changed code. The container needs to be re-created.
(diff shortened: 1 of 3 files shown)


gplssm:
I'm giving it a try? Can you tell me where have to expect the docker volume now? Can't find it in the project directory

gnn:
> I'm giving it a try? Can you tell me where have to expect the docker volume now? Can't find it in the project directory

It should be under "docker/database-data", inside `egon-data`'s working directory.

gplssm:
I found it using `docker inspect egon-data-local-database-container | grep volume`

```
"Mounts": [
    {
        "Type": "bind",
        "Source": "/home/local/RL-INSTITUT/guido.plessmann/egon-data-docker-volume/docker/database-data",
        "Destination": "/var/lib/postgresql/data",
        "Mode": "rw",
        "RW": true,
        "Propagation": "rprivate"
    }
```

It's now located under `$USER/docker/...` again, like it was before. Would it be difficult do locate it under "docker/database-data", inside `egon-data`'s working directory

gplssm:
I can report that this worked well for my case! Thanks @gnn !

Could we please merge it to dev in order to make it easier to use?

PR #240, opened 2021-04-30
Features/otg substation mapping

ulfmueller:
the subst_id is changed to bus_id in order to simplify the naming conventions throughout the process. The main function is that the mapping of the osmTGmod bus ids and the ids from the substation scripts are realized.

ToDo:
check out the mv grid district scripts and see if the id renaming has impact and requires changes.

### DIFF
--- a/src/egon/data/airflow/dags/pipeline.py
+++ b/src/egon/data/airflow/dags/pipeline.py
@@ -332,13 +332,6 @@ with airflow.DAG(
     vg250_clean_and_prepare >> hvmv_substation_extraction
     vg250_clean_and_prepare >> ehv_substation_extraction
 
-    # MV grid districts
-    define_mv_grid_districts = PythonOperator(
-        task_id="define_mv_grid_districts",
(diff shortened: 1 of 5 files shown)


ClaraBuettner:
I tested the scripts and the results look fine, the bus_id in the substation tables fits to the powerflow tables. 
The task `define_mv_grid_districts` fails due to the new column names. I fixed this by changing the subst_id to bus_id (I can also push this).
Since some substations are deleted in `osmtgmod_substation` the task `define_mv_grid_districts` has to run afterwards. So another dependency is needed, isn't it?

ulfmueller:
@ClaraBuettner thx for your review and thx @gplssm for your thumbs up on that. @ClaraBuettner can you please push your changes to this branch. I can add the dependency to the pipeline.

ClaraBuettner (approved):
I run the branch again (for Schleswig-Holstein) and the results look fine; there is one hvmv-substation per mv grid district and the ids match.

IlkaCu (approved):

Issue #15, opened 2020-10-09
Review and move 'powerflow' scripts


IlkaCu:
Scripts present in the DP-section 'powerflow' need to be review in regard to the expansion of our data model and moved to the egon-data repo. 
This will most likely include changes in the data structure as we include additional data in our model. 

- [x] ego_dp_powerflow_hv_setup.sql
- [x] ego_dp_powerflow_osmtgmod_to_pypsa.sql
- [ ] ego_dp_powerflow_electrical_neighbour.sql
- [x] ego_dp_powerflow_fix_ehv_subnetworks.sql
- [x] ego_dp_powerflow_grid_future_scenarios.sql
- [x] ego_dp_powerflow_assignment_otgid.sql
- [x] ego_dp_powerflow_assignment_unid.sql
- [x] ego_dp_powerflow_create_pp_mview.sql
- [x] ego_dp_powerflow_assignment_generator.sql
- [x] ego_dp_powerflow_assignment_load.sql
- [ ] ego_dp_powerflow_assignment_storage.sql
- [x] ego_dp_powerflow_timeseries_generator.sql
- [x] ego_dp_powerflow_griddistrict_demand.py
- [ ] ego_dp_powerflow_timeseries_demand.sql
- [ ] ego_dp_powerflow_lopf_data.sql
- [ ] ego_dp_data_check.sql

PR #246, opened 2021-05-11
Assign voltage level and bus_id to power plants

### LINKED ISSUE #174: Add MaStR Lokationen data
Information about locations of power plant facilities is needed to decide on the grid connection points

ClaraBuettner:
This branch adds voltage level and bus_ids to power plants (see #15), since I needed the mastr location data, I adjusted the mastr download, so this closes #174.

@gplssm Could you check if the location data is downloaded correctly? (I tried it and it worked, I just want to be sure to have the correct deposit_id.)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -82,6 +82,8 @@ Added
   `#214 <https://github.com/openego/eGon-data/issues/214>`_
 * Integrate weather data and renewable feedin timeseries
   `#19 <https://github.com/openego/eGon-data/issues/19>`_
+* Assign voltage level and bus_id to power plants
+  `#15 <https://github.com/openego/eGon-data/issues/15>`_
 
 .. _PR #159: https://github.com/openego/eGon-data/pull/159
 
--- a/src/egon/data/airflow/dags/pipeline.py
+++ b/src/egon/data/airflow/dags/pipeline.py
@@ -392,13 +392,14 @@ with airflow.DAG(
     )
 
     power_plant_import = PythonOperator(
-        task_id="import-power-plants",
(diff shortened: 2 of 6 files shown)


gplssm (changes requested):
Thanks for implementing this!

I scrolled through the changes and it looks plausible. 

You chose the correct deposit id.
In order to verify that the download of MaStR from Zenodo works correctly, I execute the workflow locally. It could check it, because I ran into an error: the task `create-landuse-table` fails because the openstreetmap schema isn't defined (it's a fresh DB). 

grafik

I think, this task should somehow depend on the osm downloading/importing tasks...

In addition, you should consider to write a short note in the CHANGELOG.

ClaraBuettner:
> Thanks for implementing this!
> 
> I scrolled through the changes and it looks plausible.
> 
> You chose the correct deposit id.

Thank you for checking this!
> In order to verify that the download of MaStR from Zenodo works correctly, I execute the workflow locally. It could check it, because I ran into an error: the task `create-landuse-table` fails because the openstreetmap schema isn't defined (it's a fresh DB).
> I think, this task should somehow depend on the osm downloading/importing tasks...

This tasks was added in another commit and is not related to this PR. Anyhow I could fix it in this branch, or we open a new issue/branch/PR.

IlkaCu (approved):
I tested the branch on a fresh database for 'Schleswig-Holstein' and it worked well.

gplssm:
> I tested the branch on a fresh database for 'Schleswig-Holstein' and it worked well.

Thanks! I had multiple failing, obviously due to other reasons... So, please merge!

PR #249, opened 2021-05-17
Fix format of voltage levels in hvmv substation

### LINKED ISSUE #248: Update format of voltage level of substation in Bayreuth
The substation Bayreuth Süd (https://www.openstreetmap.org/way/30065262#map=18/49.91433/11.56745) is tagged with the voltage level "110000/20000". It also looks like this in grid.egon_hvmv_substation and leads to problem in upcoming scripts (e.g. the distribution of wind parks). 
The voltage level will be adjusted for this substation inside the hvmv substation extraction.

ClaraBuettner:
Closes #248


IlkaCu (approved):
Looks good. I didn't run it, as the changes are so small. 
Please update the CHANGELOG before merging.

PR #254, opened 2021-05-19
Fix order of osmtgmod tasks

### LINKED ISSUE #253: osmTGmod fails in new environment
The task `download_osm_for_osmtgmod `needs the osmtgmod repository which is currently cloned in `run_osmtgmod`. This results in error when running egon-data in a new environment. It can be fixed by cloning the repository inside the `download_osm_for_osmtgmod` task.

ClaraBuettner:
Fixes #253 
I already tested it and it solved the problem.


ulfmueller (approved):
looks good and reasonable. I trust the changes without testing them.

PR #268, opened 2021-06-01
Increased shared memory size of docker container to 16 GB

### LINKED ISSUE #267: Shared memory of docker container too small
Right now, the shared memory size (`shm_size`) isn't specified in the [docker-compose.yml](https://github.com/openego/eGon-data/blob/2da2de55bce2522eea0fbfcd8b43fd6eb021e994/src/egon/data/airflow/docker-compose.yml). This is probably the reason why the tasks `population-in-municipalities` and `import-heat-demand` (and maybe also run_osmtgmod`) fail.

gplssm:
Fixes #267 

Note, you definitely need to stop and restart the container (`docker stop egon-data-local-database-container`). Eventually, the container needs to be deleted and re-created, use `docker rm egon-data-local-database-container`.

### DIFF
--- a/src/egon/data/airflow/docker-compose.yml
+++ b/src/egon/data/airflow/docker-compose.yml
@@ -16,3 +16,4 @@ services:
       POSTGRES_PASSWORD: {--database-password}
     volumes:
     - ./database-data/:/var/lib/postgresql/data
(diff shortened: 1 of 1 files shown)


nesnoj:
Obsolete

PR #282, opened 2021-06-11
Add airflow-port to egon-data.configuration

### LINKED ISSUE #281: Allow configuring the airflow port
When egon-data should run in parallel on the same machine, the airflow port has to be changed (otherwise I get an error like `port 8080 already in use` when I try to run `egon-data serve` for the second run). 
On the server I prevent this by manually changing my airflow port and always stash the changes. 
Updating the airflow.cfg in the working directory of egon-data is not enough because this file is overwritten with the airflow.cfg of the repository.
I would like to have the airflow port as a variable in the egon-data configuration file.

ClaraBuettner:
Closes #281 
This branch allows to configure the airflow port in the egon-data.configuration file. It is needed to run egon-data in parallel on the same machine, which is very useful on the server. 
I tested this branch and it worked fine.


AmeliaNadal:
I tested the branch successfully.

AmeliaNadal (approved):

Issue #298, opened 2021-06-23
Airflow task error after implementing versioning


nailend:
The following task error pops up after implementing versioning for the mv_gird_districts-Dataset. Similiar to https://github.com/openego/eGon-data/issues/297#issuecomment-865904336 and https://github.com/openego/eGon-data/pull/260#issuecomment-864936520
I assume all tasks in the dataset are actually completed but finish with an error anyways. Therefore no further depending task can be executed. 

```*** Reading local file: /home/local/RL-INSTITUT/julian.endres/Projekte/eGon-data/egon-data-sh/airflow/logs/egon-data-processing-pipeline/define-mv-grid-districts/2021-06-22T13:47:39.582736+00:00/1.log
[2021-06-22 15:57:42,529] {taskinstance.py:670} INFO - Dependencies all met for <TaskInstance: egon-data-processing-pipeline.define-mv-grid-districts 2021-06-22T13:47:39.582736+00:00 [queued]>
[2021-06-22 15:57:42,593] {taskinstance.py:670} INFO - Dependencies all met for <TaskInstance: egon-data-processing-pipeline.define-mv-grid-districts 2021-06-22T13:47:39.582736+00:00 [queued]>
[2021-06-22 15:57:42,594] {taskinstance.py:880} INFO - 
--------------------------------------------------------------------------------
[2021-06-22 15:57:42,594] {taskinstance.py:881} INFO - Starting attempt 1 of 1
[2021-06-22 15:57:42,594] {taskinstance.py:882} INFO - 
--------------------------------------------------------------------------------
[2021-06-22 15:57:42,616] {taskinstance.py:901} INFO - Executing <Task(MV_grid_districts (versioned)): define-mv-grid-districts> on 2021-06-22T13:47:39.582736+00:00
[2021-06-22 15:57:42,625] {standard_task_runner.py:54} INFO - Started process 867185 to run task
[2021-06-22 15:57:42,685] {standard_task_runner.py:77} INFO - Running: ['airflow', 'run', 'egon-data-processing-pipeline', 'define-mv-grid-districts', '2021-06-22T13:47:39.582736+00:00', '--job_id', '55', '--pool', 'default_pool', '--raw', '-sd', 'DAGS_FOLDER/dags/pipeline.py', '--cfg_path', '/tmp/tmp4p5o2ibd']
[2021-06-22 15:57:42,686] {standard_task_runner.py:78} INFO - Job 55: Subtask define-mv-grid-districts
[...]

PR #318, opened 2021-07-07
Improve `Dataset`s

gnn:
Allow unversioned tasks as dependencies and allow marking a `Dataset` to always run.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -33,6 +33,14 @@ Added
   returns the current configuration settings. See `PR #159`_ for more
   details.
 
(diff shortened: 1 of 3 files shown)


nailend:
I checked this new feature locally with [feature/#256 hh load area profile generator ](https://github.com/openego/eGon-data/compare/features/%23256-hh-load-area-profile-generator). Works fine and solves the task-error issue in #298  but I wasn't able to check the "always execute" feature

gnn:
Damn. I merged the branch locally last week and forgot to push... D'Oh.

nesnoj:
Concerning my comment in the today's webco @gnn : You added a note to the changelog but not the docs, could u please do so?

PR #324, opened 2021-07-12
get rid of checks to ensure clean table creation

ulfmueller:
once this task is realized the tables should be deleted and created and no checks whether they exists should make a 'clean run' possibly less clean ;)


ulfmueller:
ok, then I will close this PR. Thx for checking well - I hope my problem will no longer occur.

PR #327, opened 2021-07-14
Include project name to avoid conflicting docker containers

### LINKED ISSUE #289: Conflicting docker containers when creating a second one with `egon-data serve`
When a second container gets created with  `egon-data serve` (in a separate directory), there is a name conflict between the containers. If a different name is chosen using `egon-data --docker-container-name NEW_NAME serve`, this one doesn't come up.

ClaraBuettner:
Closes #289 
@nesnoj If you would like to have a more detailed description in the troubleshooting section, please adjust this


nesnoj (approved):
I tested with one additional instance and it seems to work. The docs look sufficient to me, cheers!
But the changelog could use a note though.

PR #369, opened 2021-08-06
fixes voronoi areas without bus_id

AmeliaNadal:


ClaraBuettner (approved):
I tested just this task in normal mode and it fixed the bug. Could you add a note to the changelog and merge this?

AmeliaNadal:
Thanks for the review @ClaraBuettner, should I close the issue also or is it better that you do it?

Issue #365, opened 2021-08-04
Dependencies of datasets ending with parallel tasks are not set correctly


ClaraBuettner:
I added the HeatSupply-dataset, which ends with three parallel tasks: 
https://github.com/openego/eGon-data/blob/26df9093bb6b1d1908bd8dbdd85c02d719cf88ea/src/egon/data/datasets/heat_supply/__init__.py#L110-L118

Another dataset (HeatEtrago) depends on the HeatSupply-dataset: 
https://github.com/openego/eGon-data/blob/26df9093bb6b1d1908bd8dbdd85c02d719cf88ea/src/egon/data/airflow/dags/pipeline.py#L617-L618

But when I set the dependency only like this, the tasks of HeatEtrago run before all tasks of HeatSupply are completed. This results into errors. So I added a dependency in the old syntax to solve this temporary. 

The same problem occurs in other datasets (on a feature-branch) which end with parallel tasks.

PR #374, opened 2021-08-06
Features/#305 migrate osm t gmod

ulfmueller:
should work. The last commit 0a19024ac8c8ba3efab266e5253a2141ac78ed73 was not tested by myself, I I remember correctly. Please check if it works and then merge :)


ClaraBuettner:
I merged the current dev branch into this one and tested it successfully. Since parallel tasks at the end of a dataset lead to wrong dependencies (see #365 ) I temporary set all tasks to run successively. 

But I noticed one problem I didn't thought about before: The etrago tables are created in a separate task which is not part of a dataset yet. The tables are re-created whenever the DAG is triggered. When osmtgmod is skipped, the data is not inserted again. So the data is missing in the etrago tables. I have an idea how to solve this problem and will merge this PR when it is fixed.

ClaraBuettner:
I solved the problem with empty eTraGo-input tables by migrating the creation of etrago-tables to a new EtragoSetup-Dataset. 
In addition, the osmTGmod repo is now cloned to the working directory according to #341 and also deleted before a new dataset-version of Osmtgmod is executed to ensure a clean run. 
I will tested this for Schleswig-Holstein and will started a run for Germany. When the run finished successfully, I can merge this.

ClaraBuettner:
This branch run successfully for Germany, so I merge it into dev

PR #380, opened 2021-08-10
Features/#321 insert gas prod

AmeliaNadal:
Unfortunatly, I have tested this branch only in test mode.


ClaraBuettner:
I have one general comment when I look at the changed files: Actually we don't want to add new datasets in the 'old' structure anymore. All new datasets should be located in the `datasets` folder and also use the` Datasets`-class. 
Could you update this? I can briefly explain what you have to do if you like.

AmeliaNadal:
Thank you for your explanation @ClaraBuettner, I changed to the suitable structure.

AmeliaNadal:
I made the changes we talked about @ClaraBuettner, could you check that everything is still working fine with them please?

ClaraBuettner (approved):
I run the branch successfully again and also checked that the data was inserted. So I think you can merge this!

PR #407, opened 2021-08-25
Add egon-prefix to db-tables

### LINKED ISSUE #349: Some tables names miss the prefix `egon_`
We agreed on using the data author or source as prefix for the tables we create, e.g. `osm_`, `vg250_` (technically speaking should be `bkg_` but nevermind) and - most important `egon_`. Furtunately, most tables are named correctly.

Some of the tables in the DB miss the prefix `egon_`:
- [x] mv_grid_districts - @nailend 
- [x] district_heating_areas - @EvaWie ?
- [x] household_electricity_profiles_in_census_cells - @nailend 
- [x] industrial_sites - @ClaraBuettner @IlkaCu ?
- [x] map_zensus_district_heating_areas - - @ClaraBuettner @IlkaCu ?

Please tick as soon as your table has been renamed. Please make sure that all downstream tasks work correctly, you may have to adapt them accordingly. In many cases, the SQLAlchemy model is used so it might be sufficient to change the original `__tablename__` in the ORM but SQL scripts do rely on the exact names.

**If you cannot make it now, please make sure that at least all tables you're up to create have the prefix.** Cheers!
We could also do the renaming as one step in the end but we have to make sure it's done prior to the publication on the **OEP**.

PS: Also, singular form should be used, so e.g. `egon_mv_grid_district` instead of `egon_mv_grid_districts`. So when you gonna change the name anyway, would be great if you could consider this too.

Related: #53

nailend:
- Added egon-prefix to table-name and used singular form.
- Changed hard-coded table names 

Fixes #349

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -237,5 +237,5 @@ Bug fixes
   `#387 <https://github.com/openego/eGon-data/issues/390>`_
 * Add xlrd and openpyxl to installation setup
   `#400 <https://github.com/openego/eGon-data/issues/400>`_
-
-
+* Add prefix egon to MV grid district tables
+  `#349 <https://github.com/openego/eGon-data/issues/349>`_
--- a/src/egon/data/datasets.yml
+++ b/src/egon/data/datasets.yml
@@ -311,7 +311,7 @@ power_plants:
       capacities: "supply.egon_scenario_capacities"
       geom_germany: "boundaries.vg250_sta_union"
       geom_federal_states: "boundaries.vg250_lan"
-      mv_grid_districts: "grid.mv_grid_districts"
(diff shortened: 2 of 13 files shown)


nesnoj (changes requested):
Your changes are correct but missing downstream dependencies as some tasks have the MVGD tables hard-coded and won't find it.
Please check all other tasks for this, e.g. there are some entries `datasets.yml` file.
And check the references in docstrings etc, as well.

Also, please don't forget to test the entire pipeline from scratch.

nailend:
Merged for the weekend run into `continuous-integration/run-everything-over-the-weekend`

nesnoj (approved):
Thx @nailend

Issue #398, opened 2021-08-18
Paths contain hard-coded slashes in some tasks


nesnoj:
Old topic, but still pending:
The following tasks make use of absolute file paths which (beside of coding standards) most likely prevents the pipeline to be run under Windows (even if users of this so-called OS is not our primary target group :penguin:):
- [x] demandregio @ClaraBuettner @IlkaCu 
- [x] geothermal @ClaraBuettner 
- [x] scenario_capacities @ClaraBuettner 
- [x] calculate_dlr @ClaraBuettner 
- [x] insert-gas-grid @AmeliaNadal 

(list might be incomplete)
Low prio but it's easy to fix and saves some pain for other users.

See @gnn's post on this topic: https://github.com/openego/eGon-data/pull/338#discussion_r673135401
Related: #59

ClaraBuettner:
All these functions use files from the data bundle. I thought a path like this: 
https://github.com/openego/eGon-data/blob/7ac8e03a289003047faee268d0d34a2d3b32efd7/src/egon/data/datasets/scenario_capacities.py#L119-L121
is a relative path because it is a sub-folder of the egon-data working directory. 
So I am a bit wondering how I should change this. Should I use `Path(.)/` instead of `os.path.join()` ?

nesnoj:
Aah sorry, "absolute" is quite misleading. It's all about the slashes in the path.

In your example above, the "data_bundle_egon_data/nep2035_version2021" part will work on Linux but will crash on Windows due to the hard-coded slash as it cannot be resolved properly (Win uses "\").
Working version:
```
 target_file = os.path.join( 
     "data_bundle_egon_data",
     "nep2035_version2021",
     scenario_config('eGon2035')['paths']['capacities']) 
```
(PS: Path is preferred if we stick to @gnn's recommendations though)

ClaraBuettner:
I replaced the paths with slashes with` Path` objects in my scripts on this branch:https://github.com/openego/eGon-data/tree/features/%23398-remove-slashes-in-paths
If you like, you can also change the other tasks in this branch.

AmeliaNadal:
I replaced the paths with slashes with Path objects in my scripts on the same branch and created a pull request (and I you as reviewer @nesnoj).

nesnoj:
Thx for your efforts!

PR #426, opened 2021-09-06
Features/#329 distribute wind offshore capacities

CarlosEpia:
This feature adds wind offshore capacities to the table "supply.egon_power_plants" based on "NEP2035_V2021_scnC2035". It was already tested on the server for full Germany.

### DIFF
--- a/src/egon/data/datasets/power_plants/__init__.py
+++ b/src/egon/data/datasets/power_plants/__init__.py
@@ -13,6 +13,7 @@ from egon.data.datasets import Dataset
 import numpy as np
 from egon.data.datasets.power_plants.pv_rooftop import pv_rooftop_per_mv_grid
 import egon.data.datasets.power_plants.wind_farms as wind_onshore
+import egon.data.datasets.power_plants.wind_offshore as wind_offshore
 import egon.data.datasets.power_plants.pv_ground_mounted as pv_ground_mounted
 Base = declarative_base()
 
@@ -45,7 +46,8 @@ class PowerPlants(Dataset):
                 insert_hydro_biomass,
                  wind_onshore.insert,
                  pv_ground_mounted.insert,
-                 pv_rooftop_per_mv_grid
+                 pv_rooftop_per_mv_grid,
+                 wind_offshore.insert               
 
             ),
         )
--- a/src/egon/data/datasets/power_plants/__init__.py
+++ b/src/egon/data/datasets/power_plants/__init__.py
@@ -39,7 +39,7 @@ class PowerPlants(Dataset):
     def __init__(self, dependencies):
         super().__init__(
             name="PowerPlants",
-            version="0.0.1",
+            version="0.0.2",
             dependencies=dependencies,
             tasks=(
                 create_tables,
--- a/src/egon/data/datasets/power_plants/wind_offshore.py
+++ b/src/egon/data/datasets/power_plants/wind_offshore.py
@@ -1,104 +1,163 @@
-
 import pandas as pd
 import geopandas as gpd
 import numpy as np
 from egon.data import db
 from shapely.geometry import Point
+from pathlib import Path
+import egon.data.config
+
 
 def insert():
     """
-    Include the off shore wind parks in egon-data
+    Include the offshore wind parks in egon-data
+    locations and installed capacities based on: NEP2035_V2021_scnC2035
+
+    Parameters
+    ----------
+    *No parameters required    
     """
-    # Connect to the data base
-    offshore_path = 'data_bundle_egon_data/nep2035_version2021/NEP2035_V2021_scnC2035.xlsx'
-    offshore = pd.read_excel(offshore_path, sheet_name= 'WInd_Offshore_NEP',
-                             usecols= ['Netzverknuepfungspunkt',
-                                       'Spannungsebene in kV',
-                                       'C 2035'])
-    offshore.dropna(subset= ['Netzverknuepfungspunkt'], inplace= True)
     
+    cfg = egon.data.config.datasets()["power_plants"]
+    offshore_path = (
+        Path(".")
+        / "data_bundle_egon_data"
+        / "nep2035_version2021"
+        / cfg["sources"]["nep_2035"]
+    )
+
+    offshore = pd.read_excel(
+        offshore_path,
+        sheet_name="WInd_Offshore_NEP",
+        usecols=["Netzverknuepfungspunkt", "Spannungsebene in kV", "C 2035"],
+    )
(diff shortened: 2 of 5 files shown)


src/egon/data/datasets/power_plants/wind_offshore.py
```diff
@@ -0,0 +1,125 @@
...
+import numpy as np
+from egon.data import db
+from shapely.geometry import Point
+
+def insert():
+    """
+    Include the off shore wind parks in egon-data
+    """
+    # Connect to the data base
+    offshore_path = 'data_bundle_egon_data/nep2035_version2021/NEP2035_V2021_scnC2035.xlsx'
```
IlkaCu, line 13: Please adjust the definition of the file path to remove hard-coded slashes, to achieve compatibility with windows systems. The topic was discussed in issue #398. 
You can use this as an example: https://github.com/openego/eGon-data/blob/e7e8762459870cafeae92d499ee22d94080c207c/src/egon/data/datasets/heat_supply/geothermal.py#L28-L32

IlkaCu: Please also add this additional source to the corresponding section in our `datasets.yml`: 
https://github.com/openego/eGon-data/blob/e7e8762459870cafeae92d499ee22d94080c207c/src/egon/data/datasets.yml#L306-L318
You can then access this entry in the datasets.yml to avoid hard coding the files' name in your code, e.g. https://github.com/openego/eGon-data/blob/212654065bb3d679a2dc1b1d8d2369ddae809674/src/egon/data/datasets/power_plants/__init__.py#L227

src/egon/data/datasets/power_plants/wind_offshore.py
```diff
@@ -0,0 +1,186 @@
...
+        axis=1,
+        inplace=True,
+    )
+
+    # convert column "bus_id" to integer
+    offshore["bus_id"] = offshore["bus_id"].apply(int)
+    offshore["voltage_level"] = offshore["voltage_level"].apply(int)
+
+    # Look for the maximum id in the table egon_power_plants
+    sql = "SELECT MAX(id) FROM supply.egon_power_plants"
```
ClaraBuettner, line 162: You can also use the target table name from the configuration file here.

src/egon/data/datasets/power_plants/wind_offshore.py
```diff
@@ -0,0 +1,186 @@
...
+        ini_id = int(max_id + 1)
+
+    offshore = gpd.GeoDataFrame(offshore, geometry="geom", crs=4326)
+
+    # write_table in egon-data database:
+    # Reset index
+    offshore.index = pd.RangeIndex(
+        start=ini_id, stop=ini_id + len(offshore), name="id"
+    )
+
```
ClaraBuettner, line 177: In order to avoid duplicate rows when a task is cleared we often try to delete existing data, e.g. existing wind_offshore_turbines, prior to inserting it. 
Could you also do this here? Here is an example: https://github.com/openego/eGon-data/blob/9185721a0ba1d060462b8836f8b13602c480e632/src/egon/data/datasets/power_plants/__init__.py#L571-L576

src/egon/data/datasets/power_plants/__init__.py
```diff
@@ -45,7 +46,8 @@ def __init__(self, dependencies):
                 insert_hydro_biomass,
                  wind_onshore.insert,
                  pv_ground_mounted.insert,
```
ClaraBuettner, line 48: Please update the version number (e.g. to 0.0.2) . Otherwise the tasks will be skipped even if it did not run with the wind offshore turbines.

src/egon/data/datasets/power_plants/wind_offshore.py
```diff
@@ -0,0 +1,186 @@
...
+        sheet_name="WInd_Offshore_NEP",
+        usecols=["Netzverknuepfungspunkt", "Spannungsebene in kV", "C 2035"],
+    )
+    offshore.dropna(subset=["Netzverknuepfungspunkt"], inplace=True)
+
+    # Import manually generated list of wind offshore farms with their connexion
+    # points (OSM_id)
+    id_bus = {
+        "Büttel": "w136034396",
+        "Heide/West": "w30622610",
```
ClaraBuettner, line 39: I'm a bit wondering why this substation is so far away from Heide. Why did you choose this osm_id?

src/egon/data/datasets/power_plants/wind_offshore.py
```diff
@@ -0,0 +1,186 @@
...
+    # write_table in egon-data database:
+    # Reset index
+    offshore.index = pd.RangeIndex(
+        start=ini_id, stop=ini_id + len(offshore), name="id"
+    )
+
+    # Insert into database
+    offshore.reset_index().to_postgis(
+        cfg["target"]["table"],
+        schema=cfg["target"]["supply"],
```
ClaraBuettner, line 181: I think you meant 'schema' instead of 'supply', didn't you?

src/egon/data/datasets/power_plants/wind_offshore.py
CarlosEpia: It is because of the required voltage level. The closest substations to Heide are "UW Husum Nord" and "Umspannwerk Kernkraftwerk Brunsbüttel" (this one was selected). They are probably the same distance from Heide. If you consider it necessary, we can agree on a closer point with a different voltage level.

src/egon/data/datasets/power_plants/wind_offshore.py
CarlosEpia: Done!

src/egon/data/datasets/power_plants/wind_offshore.py
CarlosEpia: Done.

src/egon/data/datasets/power_plants/wind_offshore.py
CarlosEpia: Done.

src/egon/data/datasets/power_plants/__init__.py
CarlosEpia: Done.

src/egon/data/datasets/power_plants/wind_offshore.py
CarlosEpia: The connection point was changed to a closer bus.

ClaraBuettner (approved):
I run this branch successfully in test mode. Did you already run it fr Germany? If yes we can also directly merge it into dev branch and the CI-branch. But there are some conflicts which have to be resolved.

PR #427, opened 2021-09-07
Use random seed from CLI parameters for CHP and society prognosis

### LINKED ISSUE #351: Random seed for reproducibility
The random module is used in several tasks. To make a further step towards reproducibility, a predefined seed should be used.
In a linear program it's sufficient to set the seed only once according to [this SO post](https://stackoverflow.com/a/11527011). But as we're running multiple tasks in parallel, the order of the drawings is not deterministic, right?

So what about defining a seed centrally in egon-data and set it in every affected script, does this sound like a good solution to you? @gnn What would be a good place to store the seed? I couldn't find an appropriate file, maybe datasets.yml but this is not the actual intention of the file.. (apart from the fact that we wanted to switch to a more decentralized concept here)

ClaraBuettner:
Closes #351 

This branch uses the random seed from the configuration file in the CHP and society prognosis function.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -193,6 +193,8 @@ Changed
   `#391 <https://github.com/openego/eGon-data/issues/391>`_
 * Update version of zenodo download
   `#397 <https://github.com/openego/eGon-data/issues/397>`_
+* Use random seed from CLI parameters for CHP and society prognosis functions
+  `#351 <https://github.com/openego/eGon-data/issues/351>`_
 
 Bug fixes
 ---------
--- a/src/egon/data/datasets/chp/small_chp.py
+++ b/src/egon/data/datasets/chp/small_chp.py
@@ -141,7 +141,7 @@ def extension_to_areas(areas, additional_capacity, existing_chp, flh, EgonChp,
     """
     session = sessionmaker(bind=db.engine())()
 
-    np.random.seed(seed=123456)
+    np.random.seed(seed=config.settings()['egon-data']['--random-seed'])
 
     # n = 0
     # Add new CHP as long as the additional capacity is not reached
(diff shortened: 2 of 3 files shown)


nesnoj (commented):
I assume u want me to review despite the fact that this PR is still in draft mode?!

Some remarks:
- is there a reason u didn't bump the CHP version no?
- apart from the seed stuff: the small CHP stuff currently seems not to be used? At least I could not find a call of the function [`extension()`](https://github.com/openego/eGon-data/blob/1ae914887cbfefdab6863fea350f0a76bc548823/src/egon/data/datasets/chp/__init__.py#L217) except for an [commented out](https://github.com/openego/eGon-data/blob/1ae914887cbfefdab6863fea350f0a76bc548823/src/egon/data/datasets/chp/__init__.py#L55) one... ?

ClaraBuettner:
> I assume u want me to review despite the fact that this PR is still in draft mode?!
> 
I'm sorry but I'm still a bit confused by the CI-branch. I wanted to mark this as ready for review when it is tested also for Germany. 

> Some remarks:
> 
>     * is there a reason u didn't bump the CHP version no?
> 
>     * apart from the seed stuff: the small CHP stuff currently seems not to be used? At least I could not find a call of the function [`extension()`](https://github.com/openego/eGon-data/blob/1ae914887cbfefdab6863fea350f0a76bc548823/src/egon/data/datasets/chp/__init__.py#L217) except for an [commented out](https://github.com/openego/eGon-data/blob/1ae914887cbfefdab6863fea350f0a76bc548823/src/egon/data/datasets/chp/__init__.py#L55) one... ?

I did not update the version number of CHP because the small chp extension which was updated is currently not used. I'm still working on it and will 'activate' it when I'm done. 
I already added the random seed for the not-used extension of small chp to not forget it later on.

nesnoj (approved):
:+1:

ClaraBuettner:
This feature was tested successfully on CI-branch for Germany, so I merge this into dev.

PR #433, opened 2021-09-08
Fixes/#432 bump mvgd version no

### LINKED ISSUE #432: Missing table "grid.egon_mv_grid_district"
All tasks depending on the mv_grid_districts failed in a local run. 
I think this was caused because the version number of the mv grid dataset was not updated in #407 . In my case, the task was skipped but I still have the old table `grid.mv_grid_district.`

nesnoj:
Fixes #432


ClaraBuettner (approved):
Thank you!

PR #416, opened 2021-08-30
Fix bug causing empty p_set columns in DB

IlkaCu:
Closes #414. 
This should solve the problem of the empty p_set column in table `demand.egon_sites_ind_load_curves`. 
Please test the code.

PR #438, opened 2021-09-09
Fix constant p_set values for industrial sites load curves

### LINKED ISSUE #429: demand.egon_sites_ind_load_curves: p_set-timeseries is constant value
While trying the function to calculate the industrial sites' p_set timeseries for DSM, we noticed that the results are timeseries with a constant value instead of 'real' timeseries. 

The relevant function is here: [https://github.com/openego/eGon-data/blob/212654065bb3d679a2dc1b1d8d2369ddae809674/src/egon/data/datasets/industry/temporal.py#L218](url)

ClaraBuettner:
Fixes #429 

@KathiEsterl @hade9443 It seems like the order of one function call needed to be changed. I tested it in the script and it was working. I was not able to test it in a egon-data run. Could one of you maybe do this? 
You can also cherry-pick this commit to your dsm-branch and test it there.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -260,3 +260,6 @@ Bug fixes
   `#349 <https://github.com/openego/eGon-data/issues/349>`_
 * Bump MV grid district version no
   `#432 <https://github.com/openego/eGon-data/issues/432>`_
+* Fix constant p_set values for industrial site load curves
+  `#429 <https://github.com/openego/eGon-data/issues/429>`_
+
--- a/src/egon/data/datasets/industry/__init__.py
+++ b/src/egon/data/datasets/industry/__init__.py
@@ -351,7 +351,7 @@ class IndustrialDemandCurves(Dataset):
     def __init__(self, dependencies):
         super().__init__(
             name="Industrial_demand_curves",
-            version="0.0.3",
+            version="0.0.4",
             dependencies=dependencies,
             tasks=(
                 create_tables,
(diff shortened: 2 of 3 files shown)


KathiEsterl:
> Fixes #429
> 
> @KathiEsterl @hade9443 It seems like the order of one function call needed to be changed. I tested it in the script and it was working. I was not able to test it in a egon-data run. Could one of you maybe do this?
> You can also cherry-pick this commit to your dsm-branch and test it there.

We will do that and let you know. Thank you, Clara!

ClaraBuettner:
It looks like #416 was merged into the CI-branch but not into the dev. I merged #416 into this branch. 
I tested it successfully, could you please also do this?

ClaraBuettner:
I had a deeper look into the code and it turns out, this is not a bug. 
We take the demand profiles for industry from demandRegio. They assign for the subsectors 17, 19, 20 and 23 the profile 'S3_WT_SA_SO' (see here: https://github.com/openego/disaggregator/blob/69df870ddac1716e44715b4319da2f833fb75d68/disaggregator/config.py#L284-L295). 
Looking in the final report from [demandRegio](https://www.ffe.de/attachments/article/735/DemandRegio_Abschlussbericht.pdf) (figure 4.13), the shape of this profile is a constant line. 
In case of industry there are only 'Betriebsschichtprofile', which means that demandRegio assumes that the machines in these large industries are always running. 
We only see this in the profiles for the sites, because the profiles for the osm areas are a combination of profiles for different subsectors. If only one of the sub sectors is assigned to one of the other profile types, there are not only constant values.

PR #443, opened 2021-09-09
Fix using datasets ending with parallel tasks as dependencies

### LINKED ISSUE #365: Dependencies of datasets ending with parallel tasks are not set correctly

gnn:
This should fix #365. It's not tested in `continuous-integration/run-everything-over-the-weekend` yet, because I also didn't get around to run the test mode on it, yet. I hope I'll manage to do so before Friday, so the changes can be teste over the weekend. But you can already have  look and tell whether any of these changes look wrong.

### DIFF
--- a/src/egon/data/datasets/__init__.py
+++ b/src/egon/data/datasets/__init__.py
@@ -135,6 +135,7 @@ class Tasks(dict):
             self.first = {task for result in results for task in result.first}
             self.last = {task for result in results for task in result.last}
             self.update(reduce(lambda d1, d2: dict(d1, **d2), results, {}))
+            self.graph = set(tasks.graph for tasks in results)
         elif isinstance(graph, tuple):
             results = [Tasks(subtasks) for subtasks in graph]
             for (left, right) in zip(results[:-1], results[1:]):
@@ -144,6 +145,7 @@ class Tasks(dict):
             self.first = results[0].first
             self.last = results[-1].last
             self.update(reduce(lambda d1, d2: dict(d1, **d2), results, {}))
+            self.graph = tuple(tasks.graph for tasks in results)
         else:
             raise (
                 TypeError(
--- a/src/egon/data/datasets/__init__.py
+++ b/src/egon/data/datasets/__init__.py
@@ -101,6 +101,12 @@ Task = Union[Callable[[], None], Operator]
 TaskGraph = Union[Task, Set["TaskGraph"], Tuple["TaskGraph", ...]]
 
 
+def prefix(o):
+    module = o.__module__
+    parent = f"{__name__}."
+    return f"{module.replace(parent, '')}." if parent in module else ""
+
+
 @dataclass
 class Tasks(dict):
     first: Set[Task]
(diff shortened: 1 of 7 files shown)


ClaraBuettner:
Looks good, thank you! I can also run it in test mode if you like. 
And I also know that there are other datasets which could use parallel tasks (society_prognosis, demandregio). Should I also change them on this branch?

gnn:
> Looks good, thank you! I can also run it in test mode if you like.

Oh, that would be really nice. I don't know whether I'll get around to it today, so if you could check that it runs in test mode, that would be really awesome.

> And I also know that there are other datasets which could use parallel tasks (society_prognosis, demandregio). Should I also change them on this branch?

I think that would be a good idea. That way we know that the fix works for all the cases. Also, there's one more change I'd like to make and then there won't be any unpleasant surprises with that one, either.

nesnoj:
> I think that would be a good idea. That way we know that the fix works for all the cases. Also, there's one more change I'd like to make and then there won't be any unpleasant surprises with that one, either.

I'll adjust the building/street extraction too. But as my branch `features/#262-preprocess-osm-buildings-amenities` is not part of `dev` yet, I'll adjust it there and merge it into the weekend branch..

ClaraBuettner:
> Oh, that would be really nice. I don't know whether I'll get around to it today, so if you could check that it runs in test mode, that would be really awesome.
> 
I just started a run and will let you now if everything worked. 


> I think that would be a good idea. That way we know that the fix works for all the cases. Also, there's one more change I'd like to make and then there won't be any unpleasant surprises with that one, either.

Ok, I will wait for your change and then commit and push the update for demandregio and society_prognosis.

ClaraBuettner:
I run into one problem (see #444) which I temporary solved with https://github.com/openego/eGon-data/pull/443/commits/29e1af37d865112853960660ed3f5fbb5657af1f  
All other tasks run successful in a clean run in test mode. So I think this can be merged into `continuous-integration/run-everything-over-the-weekend`.

gnn:
Merged into the weekend CI branch and started it.

gnn:
If I'm reading the DAG run correctly, there where no errors related to this, so I can merge it into `dev`, right?

ClaraBuettner (approved):

ClaraBuettner:
@gnn Could this be merged into `dev` or is there still something to do?

PR #455, opened 2021-09-27
Features/#436 h2 and ch4 carriers

AmeliaNadal:
Transform the carrier 'gas' into two distinct carriers 'H2' and 'CH4'.
The changes have been successfully tested in normal mode this week-end.


src/egon/data/datasets/gas_grid.py
```diff
@@ -18,6 +19,16 @@
...
 from pathlib import Path
 
+#class GasGrid(Dataset):
+#     def __init__(self, dependencies):
+#         super().__init__(
+#             name="GasGrid",
+#             version="0.0.0.dev",
+#             dependencies=dependencies,
+#             tasks=(etrago_input_data, download_data_bundle, osmtgmod_pypsa),
+#         )
```
ClaraBuettner, line 29: I'm wondering why these lines are there but commented... Do you wanted to create a `Dataset`-class for the GasGrid?

AmeliaNadal: I thought about it, but I ran into errors and because it was not the purpose of the branch, I decide to let it as it is.

ClaraBuettner (approved):

PR #467, opened 2021-10-07
Features/#73 etrago heat timeseries table

aadit879:


src/egon/data/datasets/heat_timeseries_etrago_table/hts_etrago.py
```diff
@@ -0,0 +1,119 @@
...
+              
+        next_id = next_etrago_id('load')
+        
+        bus_ts['load_id']=np.arange(len(bus_ts))+next_id
+        
+        etrago_load = pd.DataFrame(index=range(len(bus_ts)))
+        etrago_load['scn_name'] = scenario
+        etrago_load['load_id'] = bus_ts.load_id
+        etrago_load['bus'] =bus_ts.bus_id
+        etrago_load['type'] =None  ### where are these values comming from?##########*#*#*
```
ClaraBuettner, line 81: The column values `None` or `NaN` are the default values. As you already wrote in the comment it is a bit confusing were they are coming from. So I suggest to simply delete these columns from your dataframe.

ClaraBuettner:
I run this branch successfully in test-mode. Looking at the data, I was wondering about the unit of your heat time series in the etrago tables. It should be MW, but the values seemed to be too big (>100MW for one rural heat area). Could you help me?

aadit879:
> I run this branch successfully in test-mode. Looking at the data, I was wondering about the unit of your heat time series in the etrago tables. It should be MW, but the values seemed to be too big (>100MW for one rural heat area). Could you help me?

The values in the etrago tables are in MW. The high values could be due to the aggregation of the individual census cells inside the sbust_id. I just checked a random subst_id which has over 2900 zensus_population_id matched to it. Therefore for this subst_id all the individual zensus_population_id are aggregated and therefore the value could be exceeding 100MW.

Considering the district heating areas are smaller, the number of zensus_population_id undereach area_id is comparatively less and thus having the lower hourly demand values.

ClaraBuettner:
This task was finally running successful in the CI-branch! So it can be merged into `dev` .

PR #500, opened 2021-11-04
Features/#332 allocation pumped hydro

IlkaCu:
Implements the allocation of pumped hydro capacities.


IlkaCu:
Code was successfully tested on CI branch.

src/egon/data/airflow/dags/pipeline.py
```diff
@@ -457,6 +458,7 @@
             data_bundle,
             zensus_mv_grid_districts,
             district_heating_areas,
+            power_plants,
```
ClaraBuettner, line 461: The HeatSupply-Dataset doesn't have a dependency to the PowerPlants anymore. So this line could be removed.

src/egon/data/datasets/chp/__init__.py
```diff
@@ -52,6 +52,21 @@ class EgonMaStRConventinalWithoutChp(Base):
...
+            dependencies=dependencies,
+            tasks=(
+                create_tables,
+                insert_chp_egon2035,
+                assign_heat_bus,
+                # extension
+            ),
+        )
+
+
```
ClaraBuettner, line 69: These lines are also not used anymore, looks a bit like a merge problem.

src/egon/data/datasets/storages/pumped_hydro.py
```diff
@@ -0,0 +1,341 @@
...
+    unmatched: pandas.DataFrame
+        Units for which no geolocation could be identified
+
+    located : pandas.DataFrame
+        Units with a geolocation based on their city information
+
+    """
+
+    # import geolocator
+    from geopy.geocoders import Nominatim
```
ClaraBuettner, line 278: Why is this package only imported in this function and not in the beginning of the script?

src/egon/data/airflow/dags/pipeline.py
IlkaCu: resolved by removing this dependency.

src/egon/data/datasets/storages/pumped_hydro.py
IlkaCu: Good question and no answer. ;-) I moved it

PR #503, opened 2021-11-09
Features/#485 fill table egon etrago generator

CarlosEpia:
- The script "fill_etrago_gen.py"  extracts and groups the generators in supply.egon_power_plants. The grouped generators are added to the table grid.egon_etrago_generator and grid.egon_etrago_generator_timeseries. 
- The weather data related to the wind offshore generators is still not available. Therefore the column p_max_pu is filled now with the bus_id for those generators. When the data is available, the script will use it without additional modifications.
- Already tested for SH, and merged to the CI branch for being tested next weekend.


IlkaCu (approved):
Beside this one comment that an additional dependency should be set, I approve this PR

PR #512, opened 2021-11-12
Change order of demandregio tasks

### LINKED ISSUE #511: Missing CTS demands from demandregio for eGon2035
When `demandregio.insert_cts_ind_demands()` is executed the first time, there is sometimes (depending on the order of other tasks) no data for CTS for eGon2035. 
This is caused by an issue with this function call: 
https://github.com/openego/eGon-data/blob/67f5a4462258a9483df7c19e5174f578e4c0953e/src/egon/data/datasets/demandregio/__init__.py#L419-L421
When it is executed the first time directly after installing the disaggregator tool, the column names are not updated to the nuts3 codes. This is set by the parameter `use_nuts3code=True`. Since the column names are used to map the nuts3 regions with the vg250 data, no data is returned and nothing is inserted into the table `demand.egon_demandregio_cts_ind` for CTS and eGon2035. 
When the function is called a second time, everything is working correctly (which is why there is data for industry and the other scenario). 
It is also working when `demandregio.insert_household_demands()` was executed before.

ClaraBuettner:
Temporary fixes #511 by running `demandregio.insert_household_demand` before  `demandregio.insert_cts_ind_demand`

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -297,6 +297,8 @@ Bug fixes
   `#414 <https://github.com/openego/eGon-data/issues/414>`_
 * Exchange bus 0 and bus 1 in Power-to-H2 links
   `#458 <https://github.com/openego/eGon-data/issues/458>`_
+* Fix missing cts demands for eGon2035
+  `#511 <https://github.com/openego/eGon-data/issues/511>`_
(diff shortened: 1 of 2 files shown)


CarlosEpia:
It worked correctly in my local machine.

CarlosEpia (approved):

PR #518, opened 2021-11-16
Replace pandas.datetime with datetime,datetime

### LINKED ISSUE #516: datetime.datetime should be used instead of pd.datetime
The `heat_demand_timeseries.HTS` module calls the `pandas.datetime` function in some places. It should be `datetime.datetime` to avoid trouble with future versions of pandas: https://stackoverflow.com/questions/60856866/why-is-datetime-being-removed-from-pandas.

fwitte:
Resolve #516

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -303,4 +303,5 @@ Bug fixes
   `#511 <https://github.com/openego/eGon-data/issues/511>`_
 * Add `data_bundle` to `industrial_sites` task dependencies
   `#468 <https://github.com/openego/eGon-data/issues/468>`_
-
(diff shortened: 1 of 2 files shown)


IlkaCu:
Please update CHANGELOG file and increase the version number of class HeatTimeSeries.

IlkaCu (changes requested):
> Please update CHANGELOG file and increase the version number of class HeatTimeSeries.

fwitte:
Should we merge this into CI before we merge into dev?

IlkaCu:
I think it is okay to directly merge this to dev, otherwise we will face problems as soon as a later pandas version will be installed, right?

PR #478, opened 2021-10-20
Features/#474 insert hydrogen infrastructure

fwitte:
Resolve #474
Resolve #515 
Resolve #522 
Resolve #528 
Resolve #560 

<del>This branch will insert hydrogen infrastructure in the eGon2035 scenario. It includes the commits from #405 as there are interdependencies to the changes introduced there.</del> #405 was merged into dev

# ToDo
- <del>1. make sure ids are not duplicated (see #514)</del> No priority, will be handeled in #520
- <del>2. e_nom_max parameter is missing for H2 overground steel tank storage</del> optional for now
- [x] 3. e_nom_max parameter for H2 underground seems excessively large
- [x] 4. the storage potential values from the InSpEE-DS report do not differentiate between Lower Saxony/Bremen and Schleswig-Holstein/Hamburg.
- [x] 5. successfully run on CI branch
- [x] 6. move hydrogen related scripts to hydrogen_etrago #528
- [x] 7. update scripts to work with updated saline structure data (will run on CI only, if databundle is updated before the weekend!)

# 1.
With tasks running simulteanous it sometimes happens, that the tasks try to insert the same ids for for different objects into the DB, e.g. overground and underground storage.

# 2.
Define a reasonable maximum value, this can also be done at a later point in time.

# 3. 
The full size storage capacity (if the inflation around a substations fully overlays a saltstructure) should be equal in these two pictures below.

grafik
Storage capacity on the CI branch run

grafik
Storage capacity on the local run

The issue is, that the same overall storage potential (per federal state) is written to all storage sites. Cause by these lines: https://github.com/openego/eGon-data/blob/56c4540b1ccc02180b2afa2ff6b61fc089113f3e/src/egon/data/datasets/hydrogen_etrago/storage.py#L196-L198

The selection should be `.loc[rowfilter, "potential"] = ...` instead.

# 4.
Adapt selection of federal state to match the table from InSpEE-DS report

grafik Table taken from https://www.bgr.bund.de/DE/Themen/Nutzung_tieferer_Untergrund_CO2Speicherung/Downloads/InSpeeDS_TP_Bewertungskriterien.html?nn=1547442

Tested allocation to federal states on CI server, should work now.

# 5.
Resolved on 11.11.2021, will be tested in next CI run https://github.com/openego/eGon-data/pull/478#issuecomment-966284345

# 6.
[...]

PR #520, opened 2021-11-17
Add a method to prevent unique key errors

### LINKED ISSUE #514: Errors might occur in case next_etrago_id is called simultanously
As already discussed in the AP1 meeting on 20.10.2021 the next_etrago_id() might cause trouble, if tasks are executed simultaneously. The function is called from [various places in the code](https://github.com/openego/eGon-data/search?q=next_etrago_id&type=code). Unfortunately, the ids are written via different functions:

- gdf.to_postgis
- df.to_sql
- db.execute_sql

Since the last function is not called directly from a dataframe, this might prevent a uniform way of handling this issue. I will dig into it and try to come up with a solution.

fwitte:
Resolve #514

First try for method which solves the unique key problem with the next_etrago_id method (see #514). Example implementation in ch4_storages.py.

# ToDo:

- [ ] 1. Adapt all `to_postgis()` and `to_sql()` calls
- [ ] 2. (How to) handle insertion of the same ids in different scenario name(?)
- [ ] 3. Some insertions require predefined ids and should explicitly not assert new ones

## 2.
- Approach: search ids with scn_name as filter
- Maybe-Problem: This task might not even be required, as the same bus should have the same id throughout all scenarios(?)

## 3.
- Approach: Check if _id column data are available, only assert if not
- Maybe-Problem: For some data, the _id column is made the index column before writing to the db. Is this necessary? Needs discussion

# Reproduce issue

To reproduce the duplicated key issue, try checking out the #478 branch and clear the following tasks simultanously:

- insert_h2_to_ch4_to_h2
- insert power_to_h2_to_power

or clear the upstream task hydrogen_etrago.insert-hydrogen-buses. Both tasks write on the grid.egon_etrago_link table.

### DIFF
--- a/src/egon/data/datasets/ch4_storages.py
+++ b/src/egon/data/datasets/ch4_storages.py
@@ -148,16 +148,18 @@ def import_ch4_storages():
     )    
     
     # Select next id value
-    new_id = db.next_etrago_id('store')
-    
-    gas_storages_list = pd.concat([import_installed_ch4_storages(), import_ch4_grid_capacity()])
-    gas_storages_list['store_id'] = range(new_id, new_id + len(gas_storages_list))
-    
+    gas_storages_list = pd.concat(
+        [import_installed_ch4_storages(), import_ch4_grid_capacity()]
+    )
     gas_storages_list =  gas_storages_list.reset_index(drop=True)
    
     # Insert data to db
-    gas_storages_list.to_sql('egon_etrago_store',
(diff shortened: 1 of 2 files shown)


fwitte:
The variation in usages of the next_etrago_id function is quite high. Creating the function wrapper might introduce more chances of bugs happening than it actually prevents. I suggest discussing it in the near future and come do a decision.

PR #561, opened 2021-12-08
Apply hotfix for gas voronoi by filtering for carrier

### LINKED ISSUE #560: The gas voronoi creation fails with buses inserted outside of German borders
After adding the eGon100RE bus data for the European neighbors (from pypsa-eur-sec), the gas_areas.create-voronoi tasks fail. The gas voronois should be created based on the buses in Germany only, therefore we need to exclude the European neighbor buses from buses to create the voronois on. At the moment, we can do this by filtering for 'eGon2035' in the scn_name column, this will however only work temporarily.

I have two solutions in mind:
- We rename carrier of the gas buses of European neighbors to something like `H2_FR`, `H2_BE`, .... This way, we can filter for scn_name and carrier.
- After filtering the buses for scn_name we add a routine to check which of those buses are in Germany and which not.

What do you think @AmeliaNadal, @ClaraBuettner, @ulfmueller, @IlkaCu? Rocket emoji for first solution, thumbs up for second solution, comment for other solution :).

fwitte:
Resolve #560

### DIFF
--- a/src/egon/data/datasets/gas_areas.py
+++ b/src/egon/data/datasets/gas_areas.py
@@ -49,7 +49,7 @@ class EgonH2VoronoiTmp(Base):
     geom = Column(Geometry("Polygon", 4326))
 
 
-def create_CH4_voronoi():
+def create_CH4_voronoi(scn_name='eGon2035'):
     """
     Creates voronoi polygons for CH4 buses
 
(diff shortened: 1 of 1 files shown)


fwitte:
will close in favor of a different PR, that deals with similar issues in #478

PR #652, opened 2022-01-27
Change method of household type refinement

### LINKED ISSUE #651: Change method of household types refinement in hh_profiles
When refining the census household dataset, we expand from 5 to 10 household types.
This is done by applying a  percentage distribution to clusters within 5 household types to all cell in a federal state, by multiplying it.

For example: 3 * families with child * [30%,30%,30%] = 1 couple with 1 child, 1 couple with 2 children, 1 couple with 3 children.

However, in the case where the percentage distribution does not result in an int but a float, we have to apply rounding as only an integer number of profiles can be drawn further on.

This, of course, leads to a deviation both in household distribution at nuts1 level and in the total number of household. In 'SH' the total deviation in households is  '1.0012328739158585,'  means +0.12% more household, which is negligible. 

In this method, the deviation of the total number of households per cell is also within limits. Since only one cluster is divided into 3 subgroups, all others into 2 or not at all. The deviation in total household is entirely from 0.12%.

If a cluster is expanded into 2 subgroups there is always one group >50% and one group <50% which leads to the same total amount of households but always the same rounding bias for the weaker one and therefore a deviation in the household type distribution at nuts1 level.

distribution_deviation

Therefore I recommend a different method which sustains the household type distribution at nuts1 level as well as the total amount of households.

Proposal:
- draw a pool of households type within a household type cluster (e.g. ['SO','SR']) in random order but proportionate to the nuts1 distribution
- split the pool into subgroups which represent the cells with size of corresponding amount of households

nailend:
Fixes #651  . 

- [x] the branch was merged into the `continuous-integration/run-everything-over-the-weekend-v2`- branch.


nesnoj:
Now running on one of our DE instances.
I bumped only the profiles task's version no, I wonder if downstream stuff will be executed..

nesnoj:
All done, see RC

nesnoj:
> I bumped only the profiles task's version no, I wonder if downstream stuff will be executed..

After adding `.dev` to version no and re-execution all downstream tasks were executed as well including heat TS.
So I could not reproduce the putative bug @fwitte @gnn

nailend:
Checked the results of the new refinement method. Much better accuracy and also the total number of households is maintained.
grafik

nesnoj (approved):
LGTM!

PR #681, opened 2022-02-24
Features/#672 annualize capital costs

### LINKED ISSUE #672: Annualize `capital_costs` in eTraGo tables
The `capital_costs` are currently overnight costs, which might be misleading. In addition eTraGo needs annualized costs for the investment optimization, so we will annualize the capital costs directly in eGon-data.

ClaraBuettner:
Fixes #672 . 

This branch annualizes the capital_costs. Overnight investment costs, lifetime and capital_costs are stored in scenario_parameters table and accessed by the other datasets which insert data to etrago tables. 

@fwitte: I deleted the capital_cost entry for H2 feedin in 577cec2011cfc6256fbf315dd61a9f85504be456 and updated the H2_to_CH4 datasets here e64eb63e072a6f51aac6f8ce65921bb59d2f96cb . Please let me know if that's fine for you. 

- [x] the branch was merged into the `continuous-integration/run-everything-over-the-weekend-v2`- branch.


src/egon/data/datasets/scenario_parameters/parameters.py
```diff
@@ -181,10 +187,58 @@ def electricity(scenario):
...
+                costs, "battery storage", "lifetime"
+            ),
         }
+        # Insert annualized capital costs
+        # lines in EUR/km/MW/a
+        # transfermer, inverter, battery in EUR/MW/a
+        parameters["capital_cost"] = {}
+
+        for comp in parameters["overnight_cost"].keys():
+            parameters["capital_cost"][comp] = annualize_capital_costs(
```
ulfmueller, line 232: would it make sense to insert a variable for the interest rate and place its definition (although it is right now the default value) more centrally so we can run sensitivities more easy on this one?

src/egon/data/datasets/scenario_parameters/parameters.py
```diff
@@ -355,6 +440,25 @@ def heat(scenario):
...
+            "rural_water_tank": read_costs(
+                costs, "decentral water tank storage", "lifetime"
+            ),
+        }
+
+        # Insert annualized capital costs
+        parameters["capital_cost"] = {}
+
+        for comp in parameters["overnight_cost"].keys():
+            parameters["capital_cost"][comp] = annualize_capital_costs(
```
ulfmueller, line 457: so for there is no annualization for the mobility part. Would it make sense to at least add a comment or so that once that is added the same function is used?

src/egon/data/datasets/etrago_setup.py
```diff
@@ -247,16 +253,18 @@ class EgonPfHvStorage(Base):
     p_set = Column(Float(53))
     q_set = Column(Float(53))
     sign = Column(Float(53), server_default="1")
-    marginal_cost = Column(Float(53))
-    capital_cost = Column(Float(53))
-    state_of_charge_initial = Column(Float(53))
-    cyclic_state_of_charge = Column(Boolean)
+    marginal_cost = Column(Float(53), server_default="0.")
```
ulfmueller, line 256: I doubt that I like these server_default values. I got rid of them on purpose so that we actually see where we still do not have data. With these default values it might seem that we have nicely filled data if in case we don't... Especially s.th. like lifetime set to 'inf' could get very weird annualization results. Do you wanna get rid of NA values which are annoying (but maybe it's good that they annoy us ;)) or what is the idea behind this?

ulfmueller (commented):
It looks very good. I haven't tested it and will not do that this week anymore. I could do that at the beginning of next week, if you think it is good to have a second person testing the functionality.

src/egon/data/datasets/scenario_parameters/parameters.py
ClaraBuettner: Yes, it would definitely make sense, I will add the interest rate to the `global_settings`.

src/egon/data/datasets/etrago_setup.py
ClaraBuettner: We added these pypsa-default values to see all the parameters for eTraGo/PyPSA in the database. The infinite lifetime is maybe not the best example for this, but e.g. in case of `p_nom_min` or `marginal_cost` I liked the idea to directly see which data is used.

src/egon/data/datasets/scenario_parameters/parameters.py
ClaraBuettner: I did this in [1cff926](https://github.com/openego/eGon-data/pull/681/commits/1cff92673e75b5f2b7db97319c7c288f85007f44)

src/egon/data/datasets/scenario_parameters/parameters.py
ClaraBuettner: Done in [166bae3](https://github.com/openego/eGon-data/pull/681/commits/166bae32eee9326253c70eac5ec992f04ab708dd)

src/egon/data/datasets/etrago_setup.py
ulfmueller: Ah ok, I think I understand. I guess both ways have its pros and cons. I am still not very much convinced about these server_defaults. Apparently it is s.th. only used as sort of a development or debugging functionality. Maybe we can agree on that the goal for the final data set should be that no server defaults are still defining anything and that instead all parameters are explicitly defined. So may be we can add an issue to get rid of these defaults once we feel our data sets are complete? This can be seen as a little plausibility check, too... What do you think?

ulfmueller:
Thanks for addressing my comments. I am very pleased - it looks very nice! Do you want me to test it as well?

ClaraBuettner:
> Thanks for addressing my comments. I am very pleased - it looks very nice! Do you want me to test it as well?

I tested it and it was also running on the CI branch. So I think you don't have to test it again.

src/egon/data/datasets/etrago_setup.py
ClaraBuettner: I think I would like to move this discussion into our weekly meeting...

ulfmueller (approved):

PR #700, opened 2022-03-16
neighbor lines get global scenario specific lifetime parameter

ulfmueller:
Fixes # . 

- [ ] the branch was merged into the `continuous-integration/run-everything-over-the-weekend-v2`- branch.

@ClaraBuettner can you test it? - maybe after you added the eGon100RE values in the scenario parameter table (i.e.  #699 )- because like this it will throw an error for that scenario


ClaraBuettner:
> @ClaraBuettner can you test it? - maybe after you added the eGon100RE values in the scenario parameter table (i.e. #699 )- because like this it will throw an error for that scenario

I could also merge your branch into #699

ulfmueller:
Sure, con mucho gusto :+1:

PR #703, opened 2022-03-18
Share "/dev/shm" between host and container

### LINKED ISSUE #267: Shared memory of docker container too small

### LINKED ISSUE #702: PostgreSQL sometimes raises a "could not resize shared memory segment" error during a full worfklow run
This was [mentioned](https://github.com/openego/eGon-data/issues/377#issuecomment-1066445745) in issue #377 and since there was no issue for this error yet, I'd thought I'd open one in order to gather all the information about it. I found a good explanation on why this happens in [this thread][0] and I also figured out two possible solutions from the Docker docs and [this article about "Shared Memory & Docker"][4]:

  - manually configure the size of "/dev/shm" in Docker's compose file using [`shm_size`][1], or
  - make the Docker container use the host's "/dev/shm" by using a ["bind mount"][2] under the service's [`volumes`][3] key in the compose file.

I'll try the second option first, because that way I don't have to figure out to what value `shm_size` has to be set in order to prevent the error from occurring.

Btw., please correct me if the title is wrong and the error also occurs when running the workflow limited to `Schleswig-Holstein`.

[0]: https://www.postgresql.org/message-id/CAG903PrStVFsFbh02MrANe0M6wgFGwf_mzVcd%2BZeL-D57QOggQ%40mail.gmail.com
[1]: https://docs.docker.com/compose/compose-file/compose-file-v3/#shm_size
[2]: https://docs.docker.com/storage/bind-mounts/
[3]: https://docs.docker.com/compose/compose-file/compose-file-v3/#volumes
[4]: https://datawookie.dev/blog/2021/11/shared-memory-docker/

gnn:
Make the Docker container use the host's "/dev/shm" by using a ["bind mount"][2] under the service's [`volumes`][3] key in the compose file.

[2]: https://docs.docker.com/storage/bind-mounts/
[3]: https://docs.docker.com/compose/compose-file/compose-file-v3/#volumes

Checking free space on the container for the weekend run using `docker exec weekend-run-container df -h`, we get

```
Filesystem                         Size  Used Avail Use% Mounted on
shm                                 64M   16K   64M   1% /dev/shm
```

without this fix applied and

```
Filesystem                         Size  Used Avail Use% Mounted on
tmpfs                              158G   32K  158G   1% /dev/shm
```

with the commit, so this looks like it fixes #702. It should also fix #267 .

## Before merging into `dev`, please make sure that:

- [x] The `CHANGELOG.rst` was updated.
- [x] New and adjusted code is formated using `black` and `isort`.
- [x] The `Dataset`-version is updated when existing datasets are adjusted. 
- [x] The branch was merged into the `continuous-integration/run-everything-over-the-weekend-v2`-branch.
- [x] The workflow is running successful in `Schleswig-Holstein` mode.
- [x] The workflow is running successful in `Everything` mode.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -190,8 +190,10 @@ Added
 * Add efficiency, capital_cost and marginal_cost to gas related data in
   etrago tables `#596 <https://github.com/openego/eGon-data/issues/596>`_
 
-
 .. _PR #159: https://github.com/openego/eGon-data/pull/159
+.. _PR #703: https://github.com/openego/eGon-data/pull/703
+.. _#702: https://github.com/openego/eGon-data/issues/702
+.. _#267: https://github.com/openego/eGon-data/issues/267
 
 
 Changed
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -189,6 +189,15 @@ Added
   `#566 <https://github.com/openego/eGon-data/issues/566>`_
 * Add efficiency, capital_cost and marginal_cost to gas related data in
   etrago tables `#596 <https://github.com/openego/eGon-data/issues/596>`_
+* The shared memory under `"/dev/shm"` is now shared between host and
+  container. This was done because Docker has a rather tiny default for
+  the size of `"/dev/shm"` which caused random problems. Guessing what
+  size is correct is also not a good idea, so sharing between host and
+  container seems like the best option. This restricts using `egon-data`
+  with docker to Linux and MacOS, if the latter has `"/dev/shm"` but
+  seems like the best course of action for now. Done via `PR #703`_ and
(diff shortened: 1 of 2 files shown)


nesnoj (commented):
I have no ultimate way to test this as the problem occurs only from time to time. Let's observe whether it persists in the near future.

Issue #590, opened 2021-12-20
Filter the Scigrid pipeline data


AmeliaNadal:
In the [gas grid definition](https://github.com/openego/eGon-data/blob/dev/src/egon/data/datasets/gas_grid.py) the links arriving from the pipelines download of the Scigrid data should be filtered:

- [ ] Some link are disconnected from the rest of the grid. For now they are removed manually but an automatic testing would be better.
- [x] The links with length = 0 should be removed.
- [x] Some buses are disconected of the grid and should be removed.

PR #733, opened 2022-04-21
Features/#590 filter isolated gas buses

AmeliaNadal:
Fixes third point of https://github.com/openego/eGon-data/issues/590 .

### DIFF
--- a/src/egon/data/datasets/gas_grid.py
+++ b/src/egon/data/datasets/gas_grid.py
@@ -19,7 +19,8 @@ from egon.data import config, db
 from egon.data.config import settings
 from egon.data.datasets import Dataset
 from egon.data.datasets.etrago_helpers import (
-    copy_and_modify_buses, copy_and_modify_links
+    copy_and_modify_buses,
+    copy_and_modify_links,
 )
 from egon.data.datasets.electrical_neighbours import central_buses_egon100
 from egon.data.datasets.scenario_parameters import get_sector_parameters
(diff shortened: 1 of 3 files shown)


src/egon/data/datasets/gas_grid.py
```diff
@@ -660,6 +661,33 @@ def insert_gas_pipeline_list(
...
+    -------
+    None.
+    """
+    targets = config.datasets()["gas_grid"]["targets"]
+
+    db.execute_sql(
+        f"""
+        DELETE FROM {targets['buses']['schema']}.{targets['buses']['table']}
+        WHERE "carrier" = 'CH4'
+        AND scn_name = 'eGon2035'
```
CarlosEpia, line 677: Is the problem about isolated gas buses happening only in the scenario eGon2035 or does it make sense for you to use this function also for the eGon100RE scenario?

CarlosEpia (commented):
It looks good to me. Once we agree on extending the use of the new function or not to the eGon100RE, it can be merged.

src/egon/data/datasets/gas_grid.py
AmeliaNadal: The buses gas buses of eGon100RE are a copy a the gas buses of eGon2035 so we do not need to apply this function for the eGon100RE scenario.

CarlosEpia: Thanks for the clarification.

CarlosEpia (approved):
This PR can be merged.

PR #826, opened 2022-06-30
Assign cts demand to buildings

### LINKED ISSUE #671: Assign CTS demands to buildings
Similar to #435 for households, CTS demands need to be allocated to buildings.

Idea
- Use the aggregated el. CTS profile grid district level from `demand.egon_etrago_electricity_cts`, but it's probably better to use [this function](https://github.com/openego/eGon-data/blob/3983dac0743c6361ee4e023791235c6c5bbce647/src/egon/data/datasets/electricity_demand/temporal.py#L38-L114) to obtain the timeseries.
- Disaggregation to census cell:
This profile is linearly scaled down to census cell level using the annual el. demand per census cell and scenario, it is located in `demand.egon_demandregio_zensus_electricity` (sector="service").
- Disaggregation to buildings and amenities:
  - Use only
    - buildings with amenities (`openstreetmap.osm_buildings_with_amenities`) and
    - amenities (`openstreetmap.osm_amenities_not_in_buildings`) which have no building assigned. Create synthetic building for each amenity (maybe 5x5m as in HH)
  - The annual demand and cell profile is scaled down to building level by using the building's area
  - Additionally for buildings: use count of amenities (col `n_amenities_inside`) in a building for weighting the building's demand, if I remember correctly @piaulous' analyses have shown that a linear weighting may overestimate the peak load (and therefore the demand) **TODO**.
  - In cells w/ el. demand but w/o buildings, create synthetic ones. Different to the residential sector, we have no additional data on number of buildings so the demand is the only we got. A reasonable assumption is necessary (some mean value from other cells?)  **TODO**
  - How to split up demand between buildings w/ amenities and amenities? **TODO**

nailend:
Fixes #671

Additionally:

- change PK's and add indexes to `openstreetmap` tables
- change `id` to `egon_amenity_id`
- differentiate between `geom_building`, `geom_point`, `geom_amenity` and rename


nesnoj:
Please update the CHANGELOG @nailend

nailend:
Could be merged as soon as I finish the documentation. Could also do it in a separate PR...

PR #853, opened 2022-07-28
Add extendable battries only to buses representing a substation

### LINKED ISSUE #852: Add extendable batteries only at substations
Currently there are extendable batteries at every AC node in Germany which is not necessary. To reduce the complexity, they should be only added to all substation-buses.

ClaraBuettner:
Fixes #852  .


CarlosEpia (approved):
Looks good to me.

PR #870, opened 2022-08-18
Features/#857 store intermediate results of heat demand timeseries

ClaraBuettner:
Fixes #857  .
Fixes #950 

## Before merging into `dev`-branch, please make sure that

- [x] the `CHANGELOG.rst` was updated.
- [x] new and adjusted code is formated using `black` and `isort`.
- [x] the `Dataset`-version is updated when existing datasets are adjusted.
- [x] the branch was merged into the
      `continuous-integration/run-everything-over-the-weekend`-branch.
- [x] the workflow is running successful in `test mode`.
- [x] the workflow is running successful in `Everything` mode.

PR #895, opened 2022-08-30
Add cts heat demand profiles at MV substation level to db

nailend:
Hey Clara,

I would like to have these intermediate results stored in the db as well. Can we add this to your branch?

Its the cts heat demand at mv substation level. The correspondent electricity version already exists.
Its basically a copy&paste from `electricity_demand/temporal/insert_cts_load`

Cheers


ClaraBuettner:
I cherry-picked your commit in https://github.com/openego/eGon-data/pull/870/commits/50ff8311d844a4959c6327dc7d79fbc999b037ee . Initially, I only wanted to try it out but I committed it to not solve the merge conflicts twice...

Issue #807, opened 2022-06-13
Missing flake8 config


khelfen:
At the moment there is no config for `flake8`. Since black uses 88 characters by default and `flake8` uses 79 characters, this leads to incompatibility. The `.flake8` file should at least include:

```
[flake8]
max-line-length = 88
```

khelfen:
@gnn I have created a proposal and opened a PR. See #810

khelfen:
@gnn Any update here?

gnn:
Sorry, but I missed this issue. There is no configuration for flake8 because the defaults are fine. There is however a configuration for black in "pyproject.toml" because the black developers decided that this is the only place from which they read configuration settings. And that's where the maximum line length is set to 79 characters for black because that's in line with PEP8.

khelfen:
Thanks @gnn ! Actually when running `black` per `hook` or directly in the console it seems like it is not respecting the 79 char limit but still formats everything to 88 chars for me :thinking: That's why I got these incompatibilities.

Issue #770, opened 2022-05-20
Dependency issues with black and flask


khelfen:
When installing eGon-data by running `pip install -e .[dev]` I get dependecy conflicts between `black`, `click` and `flask`. I guess in the end it comes from `apache-airflow`. To resolve this for me I have to pin `black` within the `setup.py` to version `21.12b0`.

But after installation the `pre-commit` hook for black errors out for me.

![Screenshot at 2022-05-20 08-12-38](https://user-images.githubusercontent.com/46815982/169465612-8ab709e6-223b-43ec-bafa-121a520fd5e4.png)

Trying to upgrade `black` in my main env unroots the dependecy conflict as following:

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
flask 1.1.4 requires click<8.0,>=5.1, but you have click 8.1.3 which is incompatible.
flask-appbuilder 2.3.4 requires click<8,>=6.7, but you have click 8.1.3 which is incompatible.
```

khelfen:
Changing the `black` `rev` within the `.pre-commit-config.yaml` to `22.3.0` resolves the `pre-commit` issue for me.

https://github.com/openego/eGon-data/blob/c8874748c33c11f3681a6be053662d1d5ad2047b/.pre-commit-config.yaml#L21-L25

khelfen:
Adding the little suggestions to change the `pre-commit` order to run `black` before `flake8`

khelfen:
@gnn Any update here?

gnn:
Again, sorry for not following up on this for so long. The fix is actually pretty easy, but understanding the problem took a while and I didn't have the time to do the proper research until today. 

> Changing the `black` `rev` within the `.pre-commit-config.yaml` to `22.3.0` (and reinstalling the `pre-commit` envs) resolves the `pre-commit` issue for me.

This can actually be "automated" by doing `pre-commit autoupdate` within the repository, which ultimately is the correct fix. Maybe `pre-commit autoupdate --repo 'https://github.com/psf/black'` to keep the amount of moving parts minimal.

Now, understanding the issue(s) at hand was a bit more complicated because there where two separate and independent issues.
[...]

PR #927, opened 2022-09-07
Mitigate `next_etrago_id` collisions

gnn:
Don't know whether there are any open issues which this PR solves, but it should help with `next_etrago_id` collisions by generating IDs randomly in a manner which is relatively unlikely to generate collisions.

- [ ] the branch was merged into the

None of the above has happened yet, but I'll do so before the PR gets merged. I'll also assign reviewers, but since it's better to assign individual reviewers to specific commits, I'll do that later, too.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,21 +5,21 @@
 exclude: '^(\.tox|ci/templates|\.bumpversion\.cfg)(/|$)'
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v3.4.0
+    rev: v4.3.0
     hooks:
       - id: trailing-whitespace
       - id: end-of-file-fixer
       - id: debug-statements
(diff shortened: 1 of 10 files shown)


.pre-commit-config.yaml
```diff
@@ -5,21 +5,21 @@
...
     hooks:
       - id: isort
   - repo: https://gitlab.com/pycqa/flake8
-    rev: 3.9.0
+    rev: 3.9.2
     hooks:
       - id: flake8
   - repo: https://github.com/psf/black
-    rev: 20.8b1
+    rev: 22.8.0
```
nesnoj, line 22: How does this relate to #770?
Please checkout @khelfen's proposal on this #811

(Somewhat) related: #807

nesnoj:
@nesnoj

ClaraBuettner:
This problem was solved in #1287

Issue #874, opened 2022-08-23
Isolated transformers in eHV grid


ClaraBuettner:
There are some isolated transformers / eHV buses in the eHV grid which should be deleted to avoid problems in the grid optimization and clustering.

PR #966, opened 2022-10-05
Delete transformers before deleting buses

ClaraBuettner:
Deals with #874 

I deleted the components in the wrong order, so the isolated buses are still in the data model. Changing the order solves the problem. 


ClaraBuettner:
merged into `continuous-integration/run-everything-2022-11-01`

Issue #962, opened 2022-10-04
Add final buildings mapping table


nailend:
Create a mapping table for all residential and cts buildings used including 

- [x] building_id
- [x] sector
- [x] zensus_population_id
- [x] bus_id
- [x] osm (or synthetic)

Issue #684, opened 2022-03-04
Distribution of PV rooftop capacities


nesnoj:
Distribution of PV rooftop capacities for SQ and 2035 scenario. Replaces temporary method implemented in #255.

Distribution
- [x] Distribute SQ capacities from MaStR (if no coords available: use geocoding using postal code and place)
- [x] Distribute eGon2035 capacities (table `supply.egon_scenario_capacities`) using demand and/or SQ's installed capacities

Single units for distribution grid
- [ ] Write single units to table (details will follow later...)
- [ ] Generate feedin timeseries in advance or provide function to do this ad-hoc (let's discuss...)

Aggregated units for transmission grid
- [x] Write aggregated units to generator table (cf. [this code](https://github.com/openego/eGon-data/blob/891a4e9f6da3a1342055fc31e64f8b23017b1d7f/src/egon/data/datasets/power_plants/pv_rooftop.py#L204-L217))
- [x] Generate aggregated feedin timeseries and write to table (cf. [this code](https://github.com/openego/eGon-data/blob/891a4e9f6da3a1342055fc31e64f8b23017b1d7f/src/egon/data/datasets/power_plants/pv_rooftop.py#L204-L217))

PR #1008, opened 2022-10-25
Features/#988 distribute home batteries within grid districts

### LINKED ISSUE #988: Distribute home batteries within mv grid districts
Home batteries need to be distributed to consumers within the mv grid districts.

In case of the eGon2035 scenario, the total installed home battery capacity is derived from the NEP 2035 and distributed according to #749 .
I case of the eGon100RE scenario the capacity is derived from PyPSA-Eur-Sec.

Only consumers with a photovoltaic system are considered in the distribution of battery capacities. The distribution thus depends on #749 and #684 .

khelfen:
Fixes #988 . This PR depends on #684 and #962 .


src/egon/data/datasets/sanity_checks.py
```diff
@@ -769,8 +771,8 @@ def egon_power_plants_pv_roof_building():
             dataset = config.settings()["egon-data"]["--dataset-boundary"]
 
             if dataset == "Schleswig-Holstein":
-                # since the required data is missing for a SH run, it is implemented
-                # manually here
+                # since the required data is missing for a SH run, it is
+                # implemented manually here
                 total_2035 = 84070
                 sh_2035 = 2700
```
nesnoj, line 777: Can't this value be obtained from scenario data? How fragile is this, e.g. does it still work for the new MaStR data?

khelfen: Done by accessing the nep scenario csv as the value for Germany is not represented within the db in a SH run

PR #1062, opened 2022-12-07
Cache `Engine`s on a per process basis

### LINKED ISSUE #799: Postgres throws `FATAL: sorry, too many clients already`
In motorized individual travel #553 I introduced parallelization in c3542f3311d686159b9f6db74cb82716dfc2938f which also leads to parallel DB queries/commits.

As reported via email postgres sometimes throws `FATAL: sorry, too many clients already`. For testing I increased the config param `max_connections` from 100 (default) to 300 and the error now occurs only rarely (and only with many other parallel tasks). Tracking the open connections I see that my tasks use a max. of 4 connections at once. I observed that the connections in my tasks are closed over time, although I don't know how long they stay idle. So with 20 tasks (additional to the other non-emobility running tasks) the limit of 100 is easily exceeded.

Shall we increase the `max_connections` or is there a better way of handling connections?

gnn:
### TODO: Edit this

I just created this so that it exist, but don't have the time to edit the description yet. Will do so later.

Fixes #799.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,12 +14,12 @@ repos:
     rev: 5.10.1
     hooks:
       - id: isort
-  - repo: https://github.com/pycqa/flake8
-    rev: 3.9.2
-    hooks:
-      - id: flake8
   - repo: https://github.com/psf/black
     rev: 22.8.0
     hooks:
       - id: black
         language_version: python3 # Should be a command that runs python3.6+
+  - repo: https://github.com/pycqa/flake8
+    rev: 3.9.2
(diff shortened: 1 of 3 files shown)


nesnoj:
This simple solution brought the connections down to a max of 62, very nice!

eGon-data_conn_record

```Tasks:                          HP eGon2035             HP eGon100RE       eMob MIT ```

So please go ahead and merge this one @gnn :)

nesnoj:
@gnn After merging, please remove the corresponding patch from the dir on both servers

nesnoj:
\*push\*

PR #1111, opened 2023-02-23
Change eMob MIT carrier names in eMob model timeseries generation

### LINKED ISSUE #1105: Carrier naming land transport EV within grid.egon_etrago_load
@nesnoj within the table  `grid.egon_etrago_load` there are two carriers for EV at the moment. `land transport EV` and `land_transport_EV`. Is there any reasoning for that? Should we change `land transport EV` to `land_transport_EV` as at the moment it is the only carrier with whitespaces instead of underscores?

nesnoj:
Fixes #1105

Hey @ClaraBuettner and @IlkaCu, here comes the carrier renaming.

One question: I did not alter the [parameters.py](https://github.com/openego/eGon-data/blob/da196626a988f69a7a6020884a844d3248201a9a/src/egon/data/datasets/scenario_parameters/parameters.py) file as they seem to apply only to the stationary batteries - I was quickly wondering that here you use `battery storage` as well but this is only to parametrize your storages which have other carrier names, right?

One reviewer is enough. Thanks! :)


ClaraBuettner:
Looking at the code, the changes are fine. I did not tested it, maybe it is enough to test it on CI? 

And yes, in parameters.py the "battery storage" parameters are used for larger batteries in the system. The name is only used internally in egon-data, so we can probably accept the inconsistency.

nesnoj:
> Looking at the code, the changes are fine. I did not tested it, maybe it is enough to test it on CI?
> 
> And yes, in parameters.py the "battery storage" parameters are used for larger batteries in the system. The name is only used internally in egon-data, so we can probably accept the inconsistency.

Thank you, merged into CI! :)

PR #1226, opened 2025-04-16
Add optional prefix to config for DAG name

CarlosEpia:


khelfen:
I think this PR is missing an additional argument `--prefix` within the [`cli.py`](https://github.com/openego/eGon-data/blob/dev/src/egon/data/cli.py) to be operational

PR #1243, opened 2025-04-30
Try airflow imports and discard if they fail

### LINKED ISSUE #1242: [BUG] Airflow imports fail in Jupyter Lab
**Describe the bug**  
When trying to import Airflow modules in JupyterLab, the following error occurs:  
`ValueError: Unable to configure formatter 'airflow'`  
This makes it difficult to work with or edit functions that rely on Airflow within JupyterLab.

---

**To Reproduce**  
1. Open a JupyterLab notebook.
2. Attempt to import Airflow modules using:
   ```python
   from egon.data.datasets.electricity_demand import *
   ```
3. The following error appears:
   ```
   ValueError: Unable to configure formatter 'airflow'
   ```

---

**Expected behavior**  
JupyterLab should be able to import Airflow modules or fail gracefully so that module editing is still possible.

---

**Screenshots**  
_Not applicable in this case (error message already included)._

---

**Environment:**
-

---

A workaround was tested using a try-except block to avoid import errors in JupyterLab:
```python
try:
    import os
    os.environ["_AIRFLOW__AS_LIBRARY"] = "1"

    # Encapsulate Airflow imports
    from airflow.models.baseoperator import BaseOperator as Operator
    from airflow.operators.python import PythonOperator
except Exception:
    Operator = object
    PythonOperator = object
```
This prevents the error by not importing the modules if the environment doesn't support it (e.g., JupyterLab). The question is whether this workaround can be safely integrated into the codebase or if it could cause unintended side effects.

---

**Additional context**  
The main goal is to edit and test non-Airflow-related parts of the code in JupyterLab. If this workaround is not suitable for the main codebase, suggestions for an alternative approach would be appreciated.

MoritzSchloesser:
Fixes #1242 .

- [ ] the branch was merged into the

### DIFF
--- a/src/egon/data/datasets/__init__.py
+++ b/src/egon/data/datasets/__init__.py
@@ -8,8 +8,16 @@ from functools import reduce
 from typing import Callable, Iterable, Set, Tuple, Union
 import re
 
-from airflow.operators import BaseOperator as Operator
-from airflow.operators.python_operator import PythonOperator
+try:
+    import os
+    os.environ["_AIRFLOW__AS_LIBRARY"] = "1"
+
+    # Encapsulate Airflow imports
+    from airflow.models.baseoperator import BaseOperator as Operator
+    from airflow.operators.python import PythonOperator
+except Exception:
+    Operator = object
+    PythonOperator = object
 from sqlalchemy import Column, ForeignKey, Integer, String, Table, orm, tuple_
 from sqlalchemy.ext.declarative import declarative_base
 


nesnoj:
I see you're using the old eGon-data `dev`, guess that's the reason?
I've set up a new instance using the latest branch `release/v2.0.0` (see instance `airflow_bug_test/working-dir`) and it works like a charm with jupyter notebook when importing:
```
from egon.data import db
from egon.data.datasets.electricity_demand_timeseries import hh_buildings
```

PR #1255, opened 2025-05-08
Features/#960 add vom for ror and reservoir

### LINKED ISSUE #960: Add VOM for ror and reservoir
For generators of carrier 'ror' and 'reservoir' no marginal costs are assigned. To not neglect them, data about VOM need to be added in the scenario_parameters script. 
Possible source are the technology data by the Danish energy agency used in PyPSA-Eur-Sec.

CarlosEpia:
Fixes #960 .

- [X] new and adjusted code is formated using `black` and `isort`.
- [X] the `Dataset`-version is updated when existing datasets are adjusted.
- [X] the workflow is running successful in `test mode`.

### DIFF
--- a/src/egon/data/datasets/scenario_parameters/parameters.py
+++ b/src/egon/data/datasets/scenario_parameters/parameters.py
@@ -380,6 +380,8 @@ def electricity(scenario):
             "wind_offshore": read_costs(costs, "offwind", "VOM"),
             "wind_onshore": read_costs(costs, "onwind", "VOM"),
             "solar": read_costs(costs, "solar", "VOM"),
+            "run_of_river": 0,
+            "reservoir": 0,
         }
 
     elif scenario == "eGon100RE":
@@ -525,6 +527,8 @@ def electricity(scenario):
             "wind_offshore": read_costs(costs, "offwind", "VOM"),
             "wind_onshore": read_costs(costs, "onwind", "VOM"),
             "solar": read_costs(costs, "solar", "VOM"),
+            "run_of_river": 0,
+            "reservoir": 0,
         }
 
     elif scenario == "eGon2021":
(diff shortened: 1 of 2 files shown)


khelfen:
This PR is using the wrong base branch (`dev`), isn't it? Should be `release/v.2.0.0` I think?

CarlosEpia:
Right. Thanks for the observation. I already changed the base.

src/egon/data/datasets/scenario_parameters/parameters.py
```diff
@@ -380,6 +380,8 @@ def electricity(scenario):
             "wind_offshore": read_costs(costs, "offwind", "VOM"),
             "wind_onshore": read_costs(costs, "onwind", "VOM"),
             "solar": read_costs(costs, "solar", "VOM"),
+            "run_of_river": 0,
+            "reservoir": 0,
```
ClaraBuettner, line 387: Where are these values coming from? Could you provide a source?

CarlosEpia: I checked the last version of [PyPSA technology-data](https://github.com/PyPSA/technology-data/tree/master/outputs) and it is still the case, that there are no available VOM data for __ror__ and __reservoir__. Since it is a common assumption to say that the marginal costs of these technologies are 0 or near 0, I included that value in the scenario_parameters script.

I could add i.e this [report](https://www.aemo.com.au/-/media/Files/Electricity/NEM/Planning_and_Forecasting/Inputs-Assumptions-Methodologies/2019/Report-Pumped-Hydro-Cost-Modelling.pdf) which mentioned for hydro generation: "all operations and maintenance costs might be categorized as fixed rather than variable, as the cost of operation is not directly tied to the amount of energy produced"

Source added in 948c9013254f8138ccaa57c439dbbbd7e3c86896

PR #1276, opened 2025-05-21
Move installation of demandregio/disaggregator to setup.py

### LINKED ISSUE #1249: [BUG] demandregio currently requires manual installation – automatic setup broken
**Describe the bug**
The automatic installation of the `demandregio` disaggregator component within the pipeline does not work as intended. As of now, users must manually clone and install the package from a specific development branch of the `disaggregator` repository. This breaks the reproducibility of the pipeline and complicates the setup process for both new and existing users.

---

**To Reproduce**

1. Follow the standard installation and setup instructions for `eGon-data`
2. Attempt to run any task that depends on `demandregio`
3. See that the required disaggregator package is missing or fails during setup
4. Workaround requires manual installation from a custom GitHub branch:
  * `features/pandas-update-plus-master` from `openego/disaggregator`

---

**Expected behavior**
`demandregio` should be installed automatically as part of the pipeline process, with no manual intervention required.

---

* **Debugging needed**

ClaraBuettner:
Fixes #1249 .

We still need to decide to which branch the installation should point. But in general, the installation works like this. 


ClaraBuettner:
Since all branches were combined (see https://github.com/openego/eGon-data/issues/1249#issuecomment-3051456472 ), the decision on the branch is not necessary anymore. The installation points to the combined branch.

CarlosEpia (approved):
Looks good to me.

PR #1281, opened 2025-06-12
Features/#1238 update data bundle

### LINKED ISSUE #1238: Create reGon-data bundle
Imported and adapted from powerd-data https://github.com/openego/powerd-data/issues/315

We need some additional files in the data bundle: 

- cts and industry demands for eGon2035 (the current file only includes data for eGon100RE), Clara
- pypsa-eur results
- hydrogen grid

@CarlosEpia, @lenzim97 : We can collect all files that are needed here, please extend the list in case you need additional files

CarlosEpia:
Fixes #1238

### DIFF
--- a/src/egon/data/datasets.yml
+++ b/src/egon/data/datasets.yml
@@ -97,9 +97,9 @@ demandregio_installation:
 demandregio_workaround:
   source:
     cache:
-      url: "https://tubcloud.tu-berlin.de/s/dKqF6wKJnLRyDws/download/cache.zip"
+      path: "data_bundle_egon_data/demand_regio_backup/cache.zip"
     dbdump:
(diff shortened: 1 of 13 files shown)


src/egon/data/datasets.yml
```diff
@@ -567,12 +567,9 @@ data-bundle:
   sources:
     zenodo:
       deposit_id: 10226009
```
khelfen, line 569: The deposit ID needs to be updated as soon as the new data bundle is on zenodo

khelfen (approved):
Hey @CarlosEpia, thank you for your work! I can't say whether all the necessary changes have been made, but the changes you've made seem reasonable! :)

PR #1290, opened 2025-06-25
allow multiple separators for voltage levels

### LINKED ISSUE #1288: [BUG] Some power plants tasks are failing for DE
**Describe the bug**  
The tasks power_plants.wind_farms.insert and power_plants.pv_ground_mounted.insert failed with this error message:

  File "/home/regon/run-de/eGon-data/src/egon/data/datasets/power_plants/pv_ground_mounted.py", line 411, in <lambda>
    lambda x: int(x.split(";")[0])
ValueError: invalid literal for int() with base 10: '220000:110000'

It works without any problem for SH, but it looks like there are some substations in the table grid.egon_hvmv_substation that uses __:__ instead of __;__ to separate the different voltage levels.

---

**To Reproduce**  
run egon-data for Germany to create the scenario eGon100RE

---

Find our which change introduces that problem into the table grid.egon_hvmv_substation and fix it. Meanwhile allow the option to separate the different voltage levels via more than one __separator__

---

CarlosEpia:
Fixes #1288.

Already tested for Germany.

### DIFF
--- a/src/egon/data/datasets/power_plants/pv_ground_mounted.py
+++ b/src/egon/data/datasets/power_plants/pv_ground_mounted.py
@@ -1,3 +1,5 @@
+import re
+
 import geopandas as gpd
 import numpy as np
 import pandas as pd
@@ -408,7 +410,7 @@ def insert():
             )
             hvmv_substation = hvmv_substation.to_crs(3035)
             hvmv_substation["voltage"] = hvmv_substation["voltage"].apply(
-                lambda x: int(x.split(";")[0])
+                lambda x: int(re.split(";|:", x)[0])
             )
             hv_substations = hvmv_substation[
                 hvmv_substation["voltage"] >= 110000
--- a/src/egon/data/datasets/power_plants/wind_farms.py
+++ b/src/egon/data/datasets/power_plants/wind_farms.py
@@ -1,3 +1,5 @@
+import re
+
 from matplotlib import pyplot as plt
 from shapely.geometry import MultiPoint, Point
 import geopandas as gpd
@@ -370,7 +372,7 @@ def wind_power_states(
     # installed capacity is bigger than max_power_mv
     hvmv_substation = hvmv_substation.to_crs(3035)
     hvmv_substation["voltage"] = hvmv_substation["voltage"].apply(
-        lambda x: int(x.split(";")[0])
+        lambda x: int(re.split(";|:", x)[0])
     )
     hv_substations = hvmv_substation[hvmv_substation["voltage"] >= 110000]
     hv_substations = hv_substations.unary_union  # join all the hv_substations


ClaraBuettner:
Should I already review this? It is still marked as draft, but looks ready. That is why I am asking.

CarlosEpia:
Right. It is ready to be reviewed.

src/egon/data/datasets/power_plants/pv_ground_mounted.py
```diff
@@ -408,7 +410,7 @@ def adapt_grid_level(pv_pot, max_dist_hv, con):
             )
             hvmv_substation = hvmv_substation.to_crs(3035)
             hvmv_substation["voltage"] = hvmv_substation["voltage"].apply(
-                lambda x: int(x.split(";")[0])
+                lambda x: int(re.split(";|:", x)[0])
             )
             hv_substations = hvmv_substation[
                 hvmv_substation["voltage"] >= 110000
```
ClaraBuettner, line 416: This seems a bit strange for me. I thought there are only substations attached to hv (110kV) in the hvmv-substation table. Are there also some that are not attached to hv?

CarlosEpia: There are multiple voltage levels there:
image

ClaraBuettner: Mmh.. But the ones with higher voltages are allowed? 
And you only take the first entry so sth. like `60000;110000` is discarded, right? 
I think all entries in the hvmv substatsion table get an underlying mv grid district. So I am wondering if some of them don't get additional wind and pv parks.

CarlosEpia: Such case like 60000;110000 does not exist because the voltage levels are listed in order. Since in this PR I am only trying to avoid problems while separating the different voltage levels and not changing the methodology already implemented about the connection of the PV generators, if you consider it necessary, I could create a new issue that deal with possible improvements on the general method.

ClaraBuettner: Yes, it would be nice if you could open an issue for that. I will approve this PR -  Sorry for the long side comments and questions!

ClaraBuettner: And when you create the issue: it is not only the case of PV parks, the same happens for wind farms (see code below)

ClaraBuettner (approved):

src/egon/data/datasets/power_plants/pv_ground_mounted.py
CarlosEpia: Sounds good to me. The issue https://github.com/openego/eGon-data/issues/1301 was created as discussed.

PR #1291, opened 2025-06-26
Fixes/#1286 fix dataset versioning

### LINKED ISSUE #1286: [BUG] Versioning is not working anymore
**Describe the bug**  
When the pipeline is restarted, all tasks run again, even if the version numbers did not change in comparison to the previous run.

---

**To Reproduce**  
Run the pipeline twice without any changes

---

**Expected behavior**  
Only datasets in which the dataset version differs from the previous run (and their downstream dependencies) are re-executed. 

---

ClaraBuettner:
Fixes #1286 .

I tested the function when calling the parameter from the file and from CLI parameters. 
While doing that, I stumbled over the misleading documentation for creating multiple scenarios from the CLI parameters, which is why I changed that. 

I only tested it in the test case but since it is not related to any data, I think it does not require a test for Germany. 

### DIFF
--- a/src/egon/data/cli.py
+++ b/src/egon/data/cli.py
@@ -178,7 +178,11 @@ from sqlalchemy.orm import Session
     "--scenarios",
     default=["status2023", "eGon2035"],
     metavar="SCENARIOS",
-    help=("List of scenario names for which a data model shall be created."),
+    help=(
+        "Scenario name for which a data model shall be created."
+        " If you want to create multiple scenarios, set the parameter multiple"
+        " times, e.g. --scenarios eGon2035 --scenarios status2023"
+        ),
     multiple=True,
     show_default=True,
 )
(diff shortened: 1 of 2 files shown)


src/egon/data/cli.py
```diff
@@ -178,7 +178,11 @@
     "--scenarios",
     default=["status2023", "eGon2035"],
```
CarlosEpia, line 179: According to the description provided in the __help__ section some lines below, this default option (list) should change. Right?

ClaraBuettner: I tried it out, since this parameter is set to `multiple=True`, the default parameter has to be a list.

CarlosEpia: All clear.

CarlosEpia (approved):

ClaraBuettner:
Thanks for reviewing! 
Should we include this in release 2.0.0?

Issue #1283, opened 2025-06-17
[FEATURE] Unify setting source and target tables


ClaraBuettner:
**Is your feature request related to a problem? Please describe.**
Right now, it is quite complicated to find out which tables are used as source and target for which dataset. 
In my mind, it would be very helpful to have a better understanding of the dependencies to have one unified way to find source and target table names. 
In addition, it would help a lot in the modularization. 

---

**Describe the solution you'd like**
I would like to define how source and target tables are listed. So far, I can imagine two options:

- Centralized in the [datasets.yml](https://github.com/openego/eGon-data/blob/dev/src/egon/data/datasets.yml): For some datasets, this structure is already used. We could make sure that it is used in each datasets (and also probably improve the structure a bit). An advantage could be that we create a central place where source tables (and therefore dependencies) are directly visible for every dataset.

- Decentral as a part of the dataset-class: We could also add `source` and `target` fields to the dataset class. That way, the definitions are maybe more present when developing the dataset code. But we don't have one file where all dependencies are listed. 

---

**Describe alternatives you've considered**
There are also other options, or we can try to combine the two listed above.

nesnoj:
This is very likely a requirement for #1298

jh-RLI:
Yes it is a quite important requirement. Regarding the automated metadata creation and OEP-publishing, its quite important to have this transparency.
Some context: https://github.com/openego/eGon-data/issues/1177#issuecomment-3035630526

If the dataset.yml defines or the Dataset class knows which tables it uses we can collect and pass this info to OMI (which implements frictionless-py) to [describe the table schema ](https://framework.frictionlessdata.io/docs/guides/describing-data.html) and add context metadata from templates to generate the resource+table metadata. This will already result in quite good metadata quality. 

Via the dataset dependencies, we can create a datapackage and add all resources.

Currently, i plan to then use the resulting datapackage oemetadata JSON which includes all Resources which have been identified in a pipeline run to automate the OEP-publishing.

I also think both or a combination of the suggestions work. The OEMetadata info and parts if the dataset.yml might document the same things. The Metadata provides more context info and also the structure of the data which hints that the info is not redundant but it might be possible to also combine info from the metadata and datapackage.yml / Dataset class.

nesnoj:
- Update docs: Add this step to how to to add dataset

Issue #1298, opened 2025-07-03
[FEATURE] Refactor and streamline metadata creation


nesnoj:
Metadata creation should be streamlined and become more easy for users.
The current metadata files are located [here](https://github.com/openego/eGon-data/tree/a90484d179f296c7b77286718ccec1e82a9e1032/src/egon/data/metadata).

Requirements:
- Automatic creation of one metadata.json for each DB table we want to publish, naming: `db_schema.db_table.json`
- Easy-to-fill-out template (e.g. yml) for users with examples on non-inferred/non-autofill fields
- Auto-validation of metadata via OMI
- Templates for common content such as licenses, context, ... (most are already there)
- OEMetadata v2 conformant #1177
- Updated how-to for users in [readme](https://github.com/openego/eGon-data/blob/dev/CONTRIBUTING.rst#add-metadata)

Open questions:
- Where/how is defined which DB tables are to be published (and hence need metadata)? Central list?
- Do we want a metadata.json for each dataset (in the pipeline class sense)?
- Do we want a metadata.json for all results?

Depends on:
- Information on data i/o for a dataset file-based datasets
#1283 
#1304 

Further work/notes:
- Remove legacy code and other auto-created metadata ([example 1](https://github.com/openego/eGon-data/blob/0358309e024e41dbb91fe8c7dbb0bf9417601014/src/egon/data/datasets/heat_demand/__init__.py#L619), [example 2](https://github.com/openego/eGon-data/blob/a90484d179f296c7b77286718ccec1e82a9e1032/src/egon/data/datasets/emobility/motorized_individual_travel/db_classes.py#L298) , just search for "add_metadata")

Please amend @jh-RLI if needed, feel free to create sub-issues.

jh-RLI:
> Automatic creation of one metadata.json for each DB table we want to publish, naming: db_schema.db_table.json

In oemetadata v2 we have introduced a new field for "topics". The pattern of 'schema.table' is quite custom to the OEP and was mainly implemented as the OEP uses the Topics to group data tables. Currently, the topics are also created on the OEDB as DB schemas, but this approach will be updated and database schemas will be harmonized to one "datasets" schema. The topics will be moved to the application, they will still exist and group data in the Webapp but they will no longer be part of the database structure.

As data tables might be part of multiple tables the oemetadata topics field is a list.

If the pattern of schema.table is not yet implemented in the egon data code, then i would vote for separating this information as suggested.

nesnoj:
[...]

PR #1304, opened 2025-07-09
Add structure to unify setting sources and targets of Datasets

ClaraBuettner:
Deals with #1283

I implemented a first draft to add the attributes `sources` and `targets` to all `Datasets`. 
The attributes are added to every `Dataset`, but remain empty as long as they are not filled in the initialization.
There are also some functions added to `DatasetSources` and `DatasetTargets`, e.g. to easily get the schema of a specific table. 

The `sources` and `targets` are written into new columns of the table `metadata.datasets`. Submodules of the datasets can access the attributes from there to avoid importing the datasets class into the submodules (which always leads to a lot of problems with circular imports). In addition, this allows to get a quick overview of the `sources` and `targets` of all datasets. 

To be able to access the entries of `metadata.datasets` before all tasks of the dataset are executed, each dataset is first registered in the table, meaning that the information on the sources and targets is added. The version is still added after the execution of all tasks, in order to keep the versioning feature running. 

I tried the new structure for the datasets `ZensusPopulation`, `ZensusMiscellaneous` and `HeatSupply`. 
We will extend the usage in the future; those three are just meant to be examples. 

If you have any questions or comments, feel free to ask. I'm happy about any feedback!

### DIFF
--- a/src/egon/data/datasets/__init__.py
+++ b/src/egon/data/datasets/__init__.py
@@ -443,13 +443,11 @@ def load_sources_and_targets(
     Parameters
     ----------
         name (str): Name of the dataset.
-        version (str): Version of the dataset.
 
     Returns
     -------
         Tuple[DatasetSources, DatasetTargets]
     """
-
     with db.session_scope() as session:
         dataset_entry = (
             session.query(Model)
@@ -457,13 +455,14 @@ def load_sources_and_targets(
             .first()
         )
 
-    if dataset_entry is None:
-        raise ValueError(f"Dataset '{name}' not found in the database.")
+        if dataset_entry is None:
+            raise ValueError(f"Dataset '{name}' not found in the database.")
 
-    raw_sources = dataset_entry.sources or {}
-    raw_targets = dataset_entry.targets or {}
+        # Extract raw JSON dicts within the session
+        raw_sources = dict(dataset_entry.sources or {})
+        raw_targets = dict(dataset_entry.targets or {})
 
-    # Recreate DatasetSources and DatasetTargets from dictionaries
+    # Recreate objects *outside the session* (now safe)
     sources = DatasetSources(**raw_sources)
     targets = DatasetTargets(**raw_targets)
 
--- a/src/egon/data/datasets/__init__.py
+++ b/src/egon/data/datasets/__init__.py
@@ -3,15 +3,16 @@
 from __future__ import annotations
 
 from collections import abc
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from functools import partial, reduce, update_wrapper
-from typing import Callable, Iterable, Set, Tuple, Union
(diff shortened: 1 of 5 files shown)


nesnoj:
Thx a lot for this proposal @ClaraBuettner!

@jh-RLI This PR is a first draft for complementing the current dataset-class-/task-/process- based dependency graph by a data(set)-based one. It is a requirement for automatic metadata creation #1298 
Feel free to add your feedback too.

jh-RLI:
Looks very nice! 
If I understand correctly, you will then have to list all sources and targets in each Dataset class once. You can do this now as you switched the approach and during the task-/process in the pipeline only the listed datasets are used.

For me and my automation tasks regarding metadata and OEP upload, this would mean i can import a class like `HeatSupply` and get all source and target table  resources which i can then use to access the table structure from DB or pandas to generate parts of the metadata and later also to read the data to chunk upload to the OEP. 

As i see it now the sources and targets could be two datasets as input/output (n table resources) as defined dataset class like child like `HeatSupply`?  Or is the dataset always one thing which should list source and target resources.

ClaraBuettner:
> As i see it now the sources and targets could be two datasets as input/output (n table resources) as defined dataset class like child like `HeatSupply`? Or is the dataset always one thing which should list source and target resources.

Every dataset will get sources and targets which are both n tables (and files). Many datasets fill more than one table, and some tables are also filled by multiple datasets. But I'm sorry, I am not completely sure if I got your question right. Does that answer helps you?

jh-RLI:
> > As i see it now the sources and targets could be two datasets as input/output (n table resources) as defined dataset class like child like `HeatSupply`? Or is the dataset always one thing which should list source and target resources.
> 
> Every dataset will get sources and targets which are both n tables (and files). Many datasets fill more than one table, and some tables are also filled by multiple datasets. But I'm sorry, I am not completely sure if I got your question right. Does that answer helps you?

Sorry I still get started with the project ... some questions might not be fully reasonable :) Im trying to get my mental model right.

I was comparing source and target to an input and output dataset like something that would be used in processes like input -> processing -> output. 
Then i was thinking about the data publishing like what ends up on the OEP. I saw two options:
1. one dataset per process. This would be then a combination of source and process in one datapackage. 
2. Or is it reasonable to split all processes into two datasets per process for source and target?

Now i think it is more like all dataset classes represent n tables for all sources or targets. All dataset classes collect the dependency information of the complete data bundle. I think this is what i want to collect, describe with metadata and publish on the OEP.

Hope this makes it a bit more clear. Your comment also helped already.

jh-RLI:
@ClaraBuettner this PR is superseded by https://github.com/openego/eGon-data/pull/1373 isnt it?

ClaraBuettner:
> @ClaraBuettner this PR is superseded by #1373 isnt it?

Yes it is. I will close this PR as soon as the other one is merged.

PR #1322, opened 2025-07-23
Update version of isort in pre-commit-config

ClaraBuettner:
With this change, I can run the pre-commit checks again. Before I did that, I had the same problem you had @nesnoj :
```
 raise RuntimeError("The Poetry configuration is invalid:\n" + message)
          RuntimeError: The Poetry configuration is invalid:
            - data.extras.pipfile_deprecated_finder[2] must match pattern ^[a-zA-Z-_.0-9]+$
          
          [end of output]
      
      note: This error originates from a subprocess, and is likely not a problem with pip.
    error: metadata-generation-failed
    
    × Encountered error while generating package metadata.
    ╰─> See above for output.
    
    note: This is an issue with the package mentioned above, not pip.
    hint: See above for details.
Check the log at /home/clara/.cache/pre-commit/pre-commit.log
```

Poetry was installed by isort and the previously selected version caused the problem. I updated it to a newer version and it was working in my case. Could you try it out if it also works for you?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,7 +11,7 @@ repos:
       - id: end-of-file-fixer
       - id: debug-statements
   - repo: https://github.com/timothycrosley/isort
-    rev: 5.11.4
+    rev: 5.12.0
     hooks:
       - id: isort
   - repo: https://github.com/psf/black


nesnoj (approved):
Yes, this did the trick, thx!
Shall we run it on the entire codebase? -> #1326

PR #1324, opened 2025-07-23
Fixes/#1293 no h2 foreign buses

### LINKED ISSUE #1293: [BUG] Task `hydrogen_etrago.insert-h2-pipelines-for-scn` fails in eGon2035
Running current `dev` in SH mode with scenario `eGon2035`:

Task `hydrogen_etrago.insert-h2-pipelines-for-scn` fails [in this line](https://github.com/openego/eGon-data/blob/0358309e024e41dbb91fe8c7dbb0bf9417601014/src/egon/data/datasets/hydrogen_etrago/h2_grid.py#L639), message:
```
  File "/storage/428_reGon/eGon-data/SH1_v2.0.0._pre-testrun_JA/code/src/egon/data/datasets/hydrogen_etrago/__init__.py", line 228, in insert_h2_pipelines_for_scn
    insert_h2_pipelines("eGon2035")
  File "/storage/428_reGon/eGon-data/SH1_v2.0.0._pre-testrun_JA/code/src/egon/data/datasets/hydrogen_etrago/h2_grid.py", line 161, in insert_h2_pipelines
    connect_h2_grid_to_neighbour_countries(scn_name)
  File "/storage/428_reGon/eGon-data/SH1_v2.0.0._pre-testrun_JA/code/src/egon/data/datasets/hydrogen_etrago/h2_grid.py", line 649, in connect_h2_grid_to_neighbour_countries
    connection_links_df = gpd.GeoDataFrame(connection_links, geometry='geom', crs="EPSG:4326")
  File "/storage/428_reGon/eGon-data/SH1_v2.0.0._pre-testrun_JA/venv/lib/python3.8/site-packages/geopandas/geodataframe.py", line 191, in __init__
    self.set_geometry(geometry, inplace=True, crs=crs)
  File "/storage/428_reGon/eGon-data/SH1_v2.0.0._pre-testrun_JA/venv/lib/python3.8/site-packages/geopandas/geodataframe.py", line 322, in set_geometry
    raise ValueError("Unknown column %s" % col)
ValueError: Unknown column geom
```

The reason might be that `connection_links` is empty for the scenario - apparently, an H2 grid is created but it is not supposed to in `eGon2035`, right? Most likely, these settings need to be changed:

https://github.com/openego/eGon-data/blob/0358309e024e41dbb91fe8c7dbb0bf9417601014/src/egon/data/datasets/hydrogen_etrago/__init__.py#L211-L240

CarlosEpia:
Fixes #1293. Already tested in a SH run. Results available in at31 port 59732.

### DIFF
--- a/src/egon/data/datasets/pypsaeur/__init__.py
+++ b/src/egon/data/datasets/pypsaeur/__init__.py
@@ -43,7 +43,7 @@ class RunPypsaEur(Dataset):
     def __init__(self, dependencies):
         super().__init__(
             name="SolvePypsaEur",
-            version="0.0.41",
+            version="0.0.42",
             dependencies=dependencies,
             tasks=(
                 prepare_network_2,
--- a/src/egon/data/datasets/pypsaeur/__init__.py
+++ b/src/egon/data/datasets/pypsaeur/__init__.py
@@ -51,10 +51,111 @@ class RunPypsaEur(Dataset):
                 solve_network,
                 clean_database,
                 electrical_neighbours_egon100,
+                h2_neighbours_egon2035,
                 # Dropped until we decided how we deal with the H2 grid
                 # overwrite_H2_pipeline_share,
             ),
         )
+
+def countries_list():
+    return [
+        "DE",
+        "AT",
+        "CH",
+        "CZ",
+        "PL",
+        "SE",
+        "NO",
+        "DK",
+        "GB",
+        "NL",
+        "BE",
+        "FR",
+        "LU",
+    ]
+
+
+def h2_neighbours_egon2035():
+    """
+    This function load the pypsa_eur network for eGon2035, processes the H2
+    buses and insert them into the grid.egon_etrago_bus table.
+
+    Returns
+    -------
+    None.
+
+    """
+    if "eGon2035" in config.settings()["egon-data"]["--scenarios"]:
+        # Delete buses from previous executions
+        db.execute_sql(
+            """
+            DELETE FROM grid.egon_etrago_bus WHERE carrier = 'H2'
+            AND scn_name = 'eGon2035'
+            AND country <> 'DE'
+            """
+        )
+
+        # Load calculated network for eGon2035
+        target_file = (
+            Path(".")
+            / "data_bundle_egon_data"
+            / "pypsa_eur"
+            / "postnetworks"
+            / "base_s_39_lc1.25__cb40ex0-T-H-I-B-solar+p3-dist1_2035.nc"
+        )
+        n = pypsa.Network(target_file)
+
+        # Filter only H2 buses in selected foreign countries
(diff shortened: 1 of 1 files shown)


src/egon/data/datasets/pypsaeur/__init__.py
```diff
@@ -43,19 +43,120 @@ class RunPypsaEur(Dataset):
...
+            """
+        )
+
+        # Load calculated network for eGon2035
+        target_file = (
+            Path(".")
+            / "data_bundle_egon_data"
+            / "pypsa_eur"
+            / "postnetworks"
+            / "base_s_39_lc1.25__cb40ex0-T-H-I-B-solar+p3-dist1_2035.nc"
```
ClaraBuettner: Are we sure that we always have results for these intermediate scenarios?

ClaraBuettner: You could also use the read_network() function here, that is maybe easier

src/egon/data/datasets/pypsaeur/__init__.py
```diff
@@ -43,19 +43,120 @@ class RunPypsaEur(Dataset):
...
+    Returns
+    -------
+    None.
+
+    """
+    if "eGon2035" in config.settings()["egon-data"]["--scenarios"]:
+        # Delete buses from previous executions
+        db.execute_sql(
+            """
+            DELETE FROM grid.egon_etrago_bus WHERE carrier = 'H2'
```
ClaraBuettner, line 92: So just to be sure, the carrier is now "H2" instead of "H2_grid" and all scripts can work with that?

src/egon/data/datasets/pypsaeur/__init__.py
CarlosEpia: If we don’t delete those files from the data bundle, the networks should remain available.
Do you think we should create these H2 buses in a different way?

ClaraBuettner: No, it is just easier to use the function in case the path is changing. That way, you only have to change it once and not several times.

src/egon/data/datasets/pypsaeur/__init__.py
CarlosEpia: Yes. Since poWerD we use that carrier name for all H2 buses that belong to the grid.

src/egon/data/datasets/pypsaeur/__init__.py
CarlosEpia: Makes sense. Change done in f2ec2682b41f8ca3527a0dda7785d2ec4d24069a.

CarlosEpia:
The last changes were tested in a SH test. It worked as expected.

ClaraBuettner (approved):

Issue #1330, opened 2025-07-28
[BUG] pv-rooftop-to-buildings Fails when running 3 scenarios.


CarlosEpia:
The task power_plants.pv_rooftop_buildings.pv-rooftop-to-buildings Fails when creating this 3 scenarios: eGon100RE, status2019 and eGon2035. It does not fail when one eGon2035 is executed. Here the error message:

```
[2025-07-28, 13:38:13 UTC] {taskinstance.py:3313} ERROR - Task failed with exception
Traceback (most recent call last):
  File "/home/regon/run-sh/venv/lib/python3.8/site-packages/airflow/models/taskinstance.py", line 768, in _execute_task
    result = _execute_callable(context=context, **execute_callable_kwargs)
  File "/home/regon/run-sh/venv/lib/python3.8/site-packages/airflow/models/taskinstance.py", line 734, in _execute_callable
    return ExecutionCallableRunner(
  File "/home/regon/run-sh/venv/lib/python3.8/site-packages/airflow/utils/operator_helpers.py", line 252, in run
    return self.func(*args, **kwargs)
  File "/home/regon/run-sh/venv/lib/python3.8/site-packages/airflow/models/baseoperator.py", line 424, in wrapper
    return func(self, *args, **kwargs)
  File "/home/regon/run-sh/eGon-data/src/egon/data/datasets/__init__.py", line 224, in skip_task
    result = super(type(task), task).execute(*xs, **ks)
  File "/home/regon/run-sh/venv/lib/python3.8/site-packages/airflow/models/baseoperator.py", line 424, in wrapper
    return func(self, *args, **kwargs)
  File "/home/regon/run-sh/venv/lib/python3.8/site-packages/airflow/operators/python.py", line 238, in execute
    return_value = self.execute_callable()
  File "/home/regon/run-sh/venv/lib/python3.8/site-packages/airflow/operators/python.py", line 256, in execute_callable
    return runner.run(*self.op_args, **self.op_kwargs)
  File "/home/regon/run-sh/venv/lib/python3.8/site-packages/airflow/utils/operator_helpers.py", line 252, in run
    return self.func(*args, **kwargs)
  File "/home/regon/run-sh/eGon-data/src/egon/data/datasets/power_plants/pv_rooftop_buildings.py", line 2454, in pv_rooftop_to_buildings
    ) = allocate_scenarios(  # noqa: F841
[...]

PR #1338, opened 2025-08-05
Distribute pv rooftop to buildings always based on status quo distribution

ClaraBuettner:
Before, the distribution was done based on the previously created scenario. This resulted into errors when different scenarios were created (eGon2035 and eGon100RE).

Related to #1330 .


CarlosEpia (approved):
The changes make sense and it works in the SH and DE runs.

Issue #1337, opened 2025-07-31
[BUG] Some census cells are part of two district heating grids


ClaraBuettner:
**Describe the bug**  
There are a few census cells which are part of two district heating grids in one scenario. That shouldn't happen, each census cell should be only in one district heating grid.
(In the run I'm currently debugging the cells 100mN31582E41344, 100mN31582E41345, 100mN31574E41360 are effected by that. All of them are on the border of an mvgd. It might be related to that) 
The bug seems to occur sometimes and not everytime. 

This currently creates problems in the creation of heat demand profiles for district heating grids. 
In some part of the code, the cells are assigned to one grid (duplicates are dropped) and in others not. This leads to deviations in the sum of the time series and the annual values. 
For now, I will update the code in the way that the assignment to one district heating grid is consistently used. 

But in the longer term, we should find out why census cells are located in two grids and fix that.

PR #1339, opened 2025-08-05
Create distrct heating profiles with census cells selected in annual demand

ClaraBuettner:
A few cells are assigned to two district heating grids. This is a bug. To allow the correct creation of the profiles, the cells are signed to one of the grids. In the longer term, the creation of district heating areas needs to be fixed. When this is done, this commit can be reverted.

Deals with #1337 .


LennartB-dot (approved):
approved

PR #1346, opened 2025-08-12
Change name of postgres-connection type of database connection

ClaraBuettner:
The previous "pgsql" was renamed to "postgres" in airflow version 2.10

This fixes the problem in the lowflex-scenario creation


CarlosEpia (approved):
It works for SH and DE run.

Issue #1318, opened 2025-07-16
[BUG] Downloading census 2011 data in `download_zensus_*` fails due to restructured site


nesnoj:
The census website has been updated recently and the URLs need to be fixed.
According to the site, they must be obtained from the [census database](https://ergebnisse.zensus2022.de/datenbank/online/).

Workaround: copy files manually to `working-dir/zensus_population/` and set download tasks to success

nesnoj:
Related: #1240

nesnoj:
For release v2.0.0 the files will be added to the data bundle

@jh-RLI a workaround will be implemented for now, you can focus on #1240

PR #1374, opened 2025-12-04
Feature/#1240 update to zensus2022

### LINKED ISSUE #1240: Update census data to 2022
imported from powerd-data https://github.com/openego/powerd-data/issues/272

The Census 2022 data have been [released yesterday](https://www.zensus2022.de/DE/Ergebnisse-des-Zensus/_inhalt.html)  :tada: 
Consider an update

Check, if format and definition of characteristics (Merkmalsdefinition) is the same in both datasets:
- [Zensus 2011](https://www.zensus2022.de/DE/Was-ist-der-Zensus/gitterzellenbasierte_Ergebnisse_Zensus2011.html)
- [Zensus 2022](https://www.zensus2022.de/DE/Aktuelles/Hinweis_Zensusatlas.html?nn=270470)

jh-RLI:
Fixes #1240 .

This PR implements updates for the update to Zensus  2022. The Zensus data in 2022 format was converted to 2011 format to keep the code changes minimal. This conversion was done outside egon-data using several scripts. The conversion result was analyzed, results are reasonable. The analyses was done by comparing relative/absulte change, scatter plots to get an overview of the data quantity value distribution across attributes. These results are also available but not part of egon data so far.

Not all attributes from 2011 are available in zensus 2022 but all attributes which are in use currently are available. See the overview in the linked issue. If we ever need more attributes data, we can get the data from Destatis using the [contact form and requesting a "Sonderauswertung".](https://www.destatis.de/DE/Service/Kontakt/Kontakt.html)

As the data is not downloaded during pipeline run currently one must first download the databundle and replace the zensus 2011 data with 2022 data manually. At least until the  bundle is updated. Then start another run which does not download the data bundle again. To do that, i simply added an empty function in the databundle class called "skip" and replaced the download task with that function.

**The databundle will be updated once a full test run is done.**

### 📋 Pull Request Guidelines

---

## 🧑‍💻 Contributor Checklist

- [ ] Relevant **documentation is updated** (API, new features, etc.)

Optional:

- [ ] Extend the checklist for reviewers: Which aspects should be reviewed in particular?

```markdown
<!-- Example:
Please focus on validating the data handling in file XYZ.
-->
```

---

## 🔍 Reviewer Checklist

---

---

💡 **Tip:** If you add multiple reviewers, clarify who should check what — this saves time and avoids duplicated efforts.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,3 +69,6 @@ docs/_build
 
 # Mypy Cache
 .mypy_cache/
+
+# Metadata artifacts
+src/egon/data/metadata/results/converted/
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -2,4 +2,4 @@
 Authors
 =======
 
-* Jonathan Amme, Clara Büttner, Ilka Cußmann, Julian Endres, Carlos Epia, Kilian Helfenbein, Stephan Günther, Ulf Müller, Amélia Nadal, Guido Pleßmann, Francesco Witte  - https://github.com/openego/eGon-data
+* Jonathan Amme, Clara Büttner, Ilka Cußmann, Julian Endres, Carlos Epia, Kilian Helfenbein, Stephan Günther, Ulf Müller, Amélia Nadal, Guido Pleßmann, Francesco Witte, Jonas Huber  - https://github.com/openego/eGon-data
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,6 +20,8 @@ Changed
   `#1360 <https://github.com/openego/eGon-data/issues/1360>`_
 * Update setting indices in etrago tables to avoid collisions
   `#1284 <https://github.com/openego/eGon-data/issues/1284>`_
+* Update zensus data to 2022. Data required conversion to 2011 format to keep code changes minimal. Conversion was done outside of egon-data. Databundle update is WIP.
+  `#1240 <https://github.com/openego/eGon-data/issues/1240>`_
 
 Bug Fixes
 ---------
--- a/src/egon/data/datasets.yml
+++ b/src/egon/data/datasets.yml
@@ -511,7 +511,7 @@ hh_demand_profiles:
       path: "hh_el_load_profiles_100k.hdf"
 
     zensus_household_types:
-      path: "Zensus2011_Personen.csv"
+      path: "Zensus2022_Personen.csv"
 
 map_mvgrid_vg250:
   sources:
--- a/src/egon/data/datasets/electricity_demand_timeseries/hh_profiles.py
+++ b/src/egon/data/datasets/electricity_demand_timeseries/hh_profiles.py
@@ -7,7 +7,7 @@ an entire year. Spatially, the data is resolved to 100 x 100 m cells and
 provides individual and distinct time series for each household in a cell.
 The cells are defined by the dataset Zensus 2011.
 
-"""
+"""  # noqa: E501
 
 from itertools import cycle, product
 from pathlib import Path
@@ -57,7 +57,7 @@ class HouseholdElectricityProfilesInCensusCells(Base):
     subsequent data like demand profiles on MV grid level or for determining
     the peak load at load area level.
 
-    """
+    """  # noqa: E501
 
     __tablename__ = "egon_household_electricity_profile_in_census_cell"
     __table_args__ = {"schema": "demand"}
@@ -76,7 +76,7 @@ class HouseholdElectricityProfilesInCensusCells(Base):
 class EgonDestatisZensusHouseholdPerHaRefined(Base):
     """
     Class definition of table society.egon_destatis_zensus_household_per_ha_refined.
-    """
+    """  # noqa: E501
 
     __tablename__ = "egon_destatis_zensus_household_per_ha_refined"
     __table_args__ = {"schema": "society"}
@@ -96,8 +96,8 @@ class EgonEtragoElectricityHouseholds(Base):
     """
     Class definition of table demand.egon_etrago_electricity_households.
 
-    The table contains household electricity demand profiles aggregated at MV grid
-    district level in MWh.
+    The table contains household electricity demand profiles aggregated
+    at MV grid district level in MWh.
     """
 
     __tablename__ = "egon_etrago_electricity_households"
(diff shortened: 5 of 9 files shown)


jh-RLI:
I have to update some docstrings before this PR is ready for review

jh-RLI:
## Validation of the Zensus 2022 conversion

Scope: the four attributes egon-data actually reads. Everything else is documented as not consumed rather than verified.

Reference data: the Flensburg full-DE **Zensus 2011** run (`regon_dev_04-09`), read-only. Its census tables hold exactly the row counts hard-coded in the `ZensusMiscellaneous` validations, which confirms those numbers were taken from a 2011 run.

### Aggregate comparison, Germany, same cell set

| | 2011 | 2022 | Δ |
|---|---:|---:|---:|
| population | 80,324,282 | 82,570,995 | +2.80 % |
| apartments `INSGESAMT` | 40,233,016 | 42,320,834 | +5.19 % |
| buildings `INSGESAMT` | 18,449,694 | 19,024,546 | +3.12 % |
| households `INSGESAMT` | 37,324,093 | 39,539,244 | +5.94 % |

Eleven years apart, so growth of this size is expected.

The more informative check is whether combining several 2022 source tables into the 2011 shape loses mass. It does not — the category-to-total ratios are **stable or better** than 2011:

| | 2011 | 2022 |
|---|---:|---:|
| `HHGROESS_KLASS` / `INSGESAMT` | 93.9 % | 94.9 % |
| `HHTYP_FAM` / `INSGESAMT` | 95.7 % | 95.9 % |
| `HEIZTYP` / `INSGESAMT` | 94.4 % | 99.4 % |

### `quantity_q` is immaterial

Every query filters `quantity_q < 2`; the conversion sets the flag to a constant `0`. Measured on the 2011 DE run, that filter excluded **1,677 of 51,095,280** apartment rows (0.003 %) and **zero** building or household rows — so it never touched `HHGROESS_KLASS` or `HHTYP_FAM` even in 2011. The constant is a documented caveat, not a behavioural change worth acting on.

### The Personen reshape

`get_census_households_nuts1_raw()` was executed against the real `1000A-3098_de.csv` and compared with the 2011 file read by the pre-PR loader:

* identical shapes before and after `process_nuts1_census_data()` — 96 × 84 → 16 × 126
* identical column index, identical processed row index, no NaN, no negatives
* zero pattern **97.84 %** identical; 171 cells were 0 in 2011 and are populated now, and only **3** go the other way (11, 4 and 3 persons)
* persons in households 78,628,079 → 81,222,001, in both years ~98 % of total population

The documented `"."` handling never fires — the source file contains no `.` values at all.

### Cell-Key method

Zensus 2022 replaces the 2011 SAFE procedure with the [Cell-Key method](https://www.destatis.de/DE/Methoden/WISTA-Wirtschaft-und-Statistik/2024/06/geheimhaltung-cell-key-methode-zensus-062024.html) (WISTA 6/2024). Each published figure is perturbed independently, so category sums do not necessarily match the published total — the Σ ≠ `INSGESAMT` gaps are an expected property of the source, not a conversion artefact.

Destatis does not publish the perturbation parameters, but does publish quality bounds. Measured on `Zensus2022_Heizungsart_100m-Gitter.csv` over 2,628,942 cells, as the difference between the published `Insgesamt` and the sum of its six categories:

| guarantee | required | measured |
|---|---|---|
| unchanged or ±1 | ≥ 65 % | **66.9 %** |
| deviation ≥ 3 | ≤ 20 % | **17.9 %** |
| deviation ≥ 4 | ≤ 10 % | **6.2 %** |

All three hold, and this measurement combines seven perturbations against bounds written for single figures, so it is the conservative direction.

### District heating is the sensitive consumer

`district_heating_areas` computes `Fernheizung / Σ(HEIZTYP)` per 100 m cell and thresholds it at 0.3 (0.6 for status quo scenarios). This is a ratio with a threshold and no spatial aggregation, so noise does not average out.

Category codes and texts match 2011 exactly, including the literal `'Fernheizung (Fernwärme)'` the query joins on, and the source is per dwelling as in 2011. But cells qualifying at ≥ 0.3 rise from 115,909 to **194,565 (+67.9 %)** — more than the +43.9 % rise in district-heated dwellings, because a threshold amplifies. Part of that is better coverage: `HEIZTYP` now spans 99.4 % of dwellings against 94.4 % in 2011.

What this means downstream, measured rather than extrapolated:

| scenario | threshold | target share | 2011 seeded | 2022 seeded | headroom |
|---|---:|---:|---:|---:|---:|
| `reGon2037` | 0.3 | 14.00 % | 6.96 % | 9.69 % | 7.04 → **4.31 pp** |
| `status2024` | 0.6 | 8.51 % | 4.44 % | 6.27 % | 4.07 → **2.24 pp** |

The `assert diff > 0` in `district_heating_areas` therefore holds, and these are upper bounds since `area_grouping()` afterwards discards groups below the minimum demand. Headroom roughly halves, so this is the place that would break first if `district_heating_share` or `minimum_connection_rate` is retuned.

### Known deviations, stated rather than omitted

* The conversion sums the six perturbed `HEIZTYP` categories instead of using the published `Insgesamt_Heizungsart`. The Cell-Key paper advises against exactly this. The measured effect is 1.11 % in aggregate with 54.6 % of cells exact, and for a ratio drawn from one category group the summed denominator is arguably more self-consistent — but it is a conscious deviation and will be carried as a follow-up.
* The heating **energy carrier**, new in 2022, is not carried into the conversion. Also a follow-up, since adding it reopens the pre-processing chain.
* Apartments map the 2022 class `a2011bis2019` onto the 2011 `"2011 - 2019"` label as an approximation; buildings ship native MZ classes. `BAUJAHR_MZ` is not read by egon-data.

### Not covered

The full-DE run has not completed yet. The `row_count` validations will report failures until the 2022 counts are measured from that run and pinned.

jh-RLI:
Three follow-ups have been filed for work deliberately left out of this PR:

* #1529 — carry the Zensus 2022 heating **energy carrier** into the census data. New 2022 information that the conversion currently discards; adding it reopens the pre-processing chain.
* #1530 — use the published `Insgesamt` instead of summing perturbed categories. The Cell-Key method paper advises against the latter, and the `HEIZTYP` conversion does it. Measured effect 1.11 % in aggregate, so not a defect, but a conscious deviation.
* #1531 — point `data_bundle` at the new Zenodo record once the Zensus 2022 bundle is published, and consider an assertion that turns a wrong bundle into an immediate failure instead of a silent 2011 load.

Also in this PR since the review comment above: `ZensusMiscellaneous.sources` no longer claims the census grid files are downloaded from `zensus2011.de`. Those three URLs were never read and are the ones reported broken in #1318; the files come from the data bundle and are now declared that way.

PR #1439, opened 2026-07-10
Features/#1438 adapt dataset electricity neighbors to new scenarios

### LINKED ISSUE #1438: [FEATURE] adapt dataset Electricity Neighbors  to new scenarios
As part of the new version of egon-data each task of the taskgroup “Electricity_Neighbors ” should be modified to include the scenarios reGon2037, reGon2045 and status2024 (updated version of status2023 which also replaces status2019) while discarding the scenarios eGon100RE and status2019 and keeping eGon2035. This changes must also include all required to follow the new design concept where all tasks are suppose to be executed for all required scenarios (defined in egon-data.configuration.yaml –scenarios) and decide inside the task if something should be executed.

This dataset will be discarded/hardly modified once the new grid data based on Pypsa (replacement of OSMTGMOD) is available.

CarlosEpia:
Fixes #1438 .

Tested for SH. Working correctly.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -35,6 +35,13 @@ Changed
   heating areas and heat demand timeseries to the configured scenarios,
   and remove obsolete status2019/eGon100RE handling
   `#1413 <https://github.com/openego/eGon-data/issues/1413>`_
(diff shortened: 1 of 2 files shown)


LennartB-dot (approved):
seems valid!

PR #1440, opened 2026-07-10
Features/#1424 changes in taskgroup electricity flexibility for re gon scenarios

### LINKED ISSUE #1424: [FEATURE] changes in taskgroup Electricity_flexibility for reGon scenarios
Adapt the taskgroup Electricity_flexibility so it can create the data for reGon2037, reGon2045, eGon2035 and status2024. This changes must also include all require to follow the new design concept where all tasks are suppose to be executed for all required scenarios and decide inside if something should be executed.
In addition, some tests are suppose to be run so the results are verified.

CarlosEpia:
Fixes #1424 .

Tested for SH. Results look good.


LennartB-dot (approved):
all good!

PR #1472, opened 2026-08-06
Skip CHP extension areas that cannot be assigned to a bus

### LINKED ISSUE #1468: [BUG] chp.extension-NS crashes when an industrial area falls in a gap between MV grid districts
The combined_heat_and_power.chp.extension-<STATE> tasks call assign_bus_id() ([power_plants/__init__.py](vscode-webview://1r2gapo6jrosgav1kjomr19b22jt1ugp22p47ghvpjh7oc09gdb5/src/egon/data/datasets/power_plants/__init__.py)) to attach a bus_id to each candidate industrial area via a spatial join (gpd.sjoin, intersects only) with grid.egon_mv_grid_district. If an area's representative point (ST_PointOnSurface) doesn't fall inside any MV grid district polygon, bus_id stays NaN and the function's hard assertion fails, crashing the whole federal-state task — even though a single unmatched area shouldn't block the rest of the extension.

To Reproduce

1. Run the pipeline through combined_heat_and_power.chp.extension-NS (Niedersachsen).
 
2. Task fails with:AssertionError: Some power plants are not attached to a bus: 
demand   osm_id  ... voltage_level bus_id
298  309.181547  9532489  ...           7.0    NaN

3. osm_id=9532489 is "Rhenus Bulk Terminal Wilhelmshaven" (openstreetmap.osm_landuse), a harbor industrial site. Its point (8.15625°E, 53.58327°N) doesn't intersect any MV grid district polygon — the two nearest districts (bus_id 39479, 39651) are only ~125 m and ~136 m away. Likely a small gap/water area between adjacent district polygons, not genuine lack of grid coverage.

Expected behavior
Areas that can be assigned to any bus should be discarded.

CarlosEpia:
Fixes #1468. Already tested for DE.

### DIFF
--- a/src/egon/data/datasets/chp/__init__.py
+++ b/src/egon/data/datasets/chp/__init__.py
@@ -770,7 +770,7 @@ class Chp(Dataset):
     #:
     name: str = "Chp"
     #:
-    version: str = "0.0.13"
+    version: str = "0.0.14"
 
     def __init__(self, dependencies):
         super().__init__(
--- a/src/egon/data/datasets/chp/small_chp.py
+++ b/src/egon/data/datasets/chp/small_chp.py
@@ -277,9 +277,24 @@ def extension_to_areas(
                 # Assign bus_id
                 selected_areas["voltage_level"] = selected_chp["voltage_level"]
 
-                selected_areas.loc[:, "bus_id"] = assign_bus_id(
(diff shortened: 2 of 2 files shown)


lenzim97 (approved):
Looks fine for me

Issue #1437, opened 2026-07-09
[FEATURE] databundle for new scenarios


CarlosEpia:
Since the reGon2037 and reGon2045 scenarios are partially based on new data, some new input files will be required for the pipeline execution.
A copy of the current databundle is available on the project's RLI SharePoint. It will be used to collect the new files and remove the ones no longer needed. This folder will be uploaded once all code changes are complete and all scenarios have been tested.
Because the new databundle is still under development, it won't be available online as part of the DataBundle dataset. Each developer will therefore need to copy the entire folder below into their working directory:

**428_General/02-project_contents/AP 3 - Data updating and modularization/New_databundle_2026**

in your working directory.

In addition, please add a comment to this issue listing, for every file you need in the new version of the databundle: its name, its location within the folder, and a short description.

CarlosEpia:
Cached files necessary for the new scenarios when using demand-regio:
5a7ed9043e53ea4ecc0f4342b0561bcf21cded8f.csv
921fae62fa64660e8e828f07d11395d8fbc7b1d7.csv
c0940e0afe260b703004fc779f10f6c076c4bd98.csv
c1ac6d452338dad143a5c366befb1757962dc6aa.csv

This files are located in:
` demandregio-disaggregator/disaggregator/disaggregator/data_in/__cache__/ `

LennartB-dot:
Files for #1415 include the new Kraftwerksliste and updated capacity-sheet. 
It is a folder containing two xlsx-files: the new capacities and the Kraftwerksliste from the NEP2025.
The folder called "nep2037_version2025" needs to be inserted into the data_bundle.

nesnoj:
> Cached files necessary for the new scenarios when using demand-regio: 5a7ed9043e53ea4ecc0f4342b0561bcf21cded8f.csv 921fae62fa64660e8e828f07d11395d8fbc7b1d7.csv c0940e0afe260b703004fc779f10f6c076c4bd98.csv c1ac6d452338dad143a5c366befb1757962dc6aa.csv
> 
> This files are located in: `demandregio-disaggregator/disaggregator/disaggregator/data_in/__cache__/`

Hey @CarlosEpia, could you tell me where you've put the csv files? I cannot find them on SP in New_databundle_2026.. (only the old cache.zip). Thx!

lenzim97:
Updated potential areas pv_road_and_railway files are available here : 
[...]

PR #1473, opened 2026-08-06
Add electric HGV charging demand model (N2/N3/N3S)

### LINKED ISSUE #1436: [FEATURE] Add electric HGV demand model (vehicle classes N2+N3) (NEP 2025 assumes full electrification, current scenarios use hydrogen)
**Is your feature request related to a problem? Please describe.**
NEP 2025 assumes heavy goods vehicles (HGV) are fully electrified in its
future scenarios. In the currently available scenarios `eGon2037` and
`eGon100RE`, HGVs are instead modeled as hydrogen fuel-cell vehicles via the
`HeavyDutyTransport` dataset (`carrier: H2_hgv_load`, `fcev_share: 1.0`).
There is no electric HGV demand model in the pipeline, so egon-data cannot
reproduce the NEP 2025 electrification assumption.

---

**Describe the solution you'd like**
Add an electric HGV demand model to the pipeline, covering the NEP-2025-based
scenarios, or in place of the existing hydrogen-based
`HeavyDutyTransport` dataset where applicable. This includes wiring the model
into a `Dataset` class and the task graph, adding the required `datasets.yml`
configuration, and validating results against the existing
`grid.egon_etrago_load(_timeseries)` tables.

MoritzSchloesser:
Closes #1436

## What this does

Adds `HGVCharging`, a demand model for **fully electrified** heavy goods
vehicles, for the NEP-2025-aligned scenarios (`reGon2037`, `reGon2045`).
Until now the pipeline only modelled HGVs as hydrogen FCEVs via
`HeavyDutyTransport` (`carrier: H2_hgv_load`, `fcev_share: 1.0`), so no
electric HGV demand existed at all.

The two fuel technologies are modelled as **sibling datasets in the same
`mobility_demand` TaskGroup**, not merged:

| dataset | technology | scenarios |
|---|---|---|
| `HGVCharging` (new) | electric (BEV) | `reGon2037`, `reGon2045` |
| `HeavyDutyTransport` (existing) | hydrogen (FCEV) | `eGon2035` |

They were kept separate because their dependencies are almost disjoint
(`mv_grid_districts` + `osmtgmod` for the electric grid side vs.
`create_gas_polygons` for the H2 side) and their scenario coverage does not
overlap, so coupling their scheduling would add risk without benefit.

## Model outline

Charging demand is built in two stages. Stage A (site selection, energy
distribution and charging-event generation) is developed outside this repo;
this PR contains the pipeline-side integration that consumes it.

* **Charging sites** — depots (OSM logistics/retail/port land use, filtered
  to `--dataset-boundary`) and highway sites (BAST Lkw-Ladenetz locations).
* **Energy distribution** — the national scenario total per vehicle class
  (N2, N3, N3S) is distributed across sites by an area power law,
  `E[i] ∝ w[category] · A[i]^0.7`, with per-category weights fitted to
  target vehicle counts per day.
* **Charging events** — one event per vehicle visit, assigned to a physical
  charging slot; each slot's chain is gap-filled with synthetic *vacant*
  events so every slot has a continuous year-long trajectory.
* **eTraGo integration** — per MV grid district (and per site for HV/EHV),
  an aggregated battery flex model: `Bus` (`HGV_charger`) + `Link`
  (`HGV_charger`, `p_max_pu` = plugged-in capacity) + `Store`
  (`battery_storage`, `e_min_pu`/`e_max_pu` = SOC band) + `Load`
  (`land_transport_HGV`, `p_set` = driving load). Highway **daytime** (MCS)
  charging carries no vehicle-event concept and is written as a plain
  profile-based fixed load instead.

This mirrors the MIV (`MotorizedIndividualTravel`) model structure
deliberately, including the `load_time_series` (dumb charging) vs.
`driving_load_time_series` split and the lowflex-scenario convention.

### Model documentation

The model is documented in full in the companion `HGV/` repository, which
holds the Stage A implementation and the standalone Stage B script:

* `DOCUMENTATION.md` (~800 lines) — architecture, both Stage A steps, the
  four database tables, spatial assignment, eTraGo integration, the flex
  model, reference tables for vehicle classes / depot categories / charger
  ratings, and a "Known limitations" section.
* `CONTEXT.md` — domain glossary (charging event, vacant event, physical
  slot, vehicles per day, depot vs. highway site), kept aligned with the MIV
  vocabulary.
* `docs/adr/0001-per-event-full-year-vehicle-trajectories.md` and
  `docs/adr/0002-slot-chained-vehicle-trajectories.md` — the two architecture
  decision records behind the SOC model. ADR-0002 supersedes ADR-0001 and
  documents why trajectories are chained per *physical slot* rather than per
  charging session (per-session trajectories let ~365 idle sessions swamp the
  one active session at a busy multi-slot charging point, flattening the SOC
  band).

Reviewers wanting the modelling rationale rather than the integration code
should start with ADR-0002 and `DOCUMENTATION.md` §10 (flex model).

## Changes

```
 docs/data/mobility_demand.rst                              |   8 +
 src/egon/data/airflow/dags/pipeline.py                      |  17 +-
 src/egon/data/datasets.yml                                  |  48 +-
 src/egon/data/datasets/__init__.py                          |  12 +-
 src/egon/data/datasets/electrical_neighbours.py             |   8 +
 .../emobility/heavy_duty_transport/__init__.py              |   9 +-
 .../datasets/emobility/hgv_charging/__init__.py             | 111 ++++
 .../datasets/emobility/hgv_charging/db_classes.py           | 142 +++++
 .../emobility/hgv_charging/etrago_integration.py            | 597 +++++++++++++
 .../datasets/emobility/hgv_charging/fill_tables.py          | 194 +++++++
 .../datasets/emobility/hgv_charging/scenarios.py            |  19 +
 .../emobility/hgv_charging/spatial_assignment.py            | 115 ++++
 .../motorized_individual_travel/model_timeseries.py         |  12 +-
 src/egon/data/datasets/sanity_checks.py                      |  11 +
```

New tables (`demand` schema): `egon_hgv_charging_site`,
`egon_hgv_charging_point`, `egon_hgv_charging_event`, `egon_hgv_profile`.

Extended: `grid.egon_etrago_bus`, `_link(_timeseries)`,
`_store(_timeseries)`, `_load(_timeseries)`.

### Input data

The precomputed Stage A output ships in the **data bundle**, one subfolder per
scenario holding `sites.gpkg`, `charging_points.csv`, `charging_events.csv`
and `profiles.csv`:

```
data_bundle_egon_data/hgv_charging/<scenario>/
```

`hgv_input_dir` in *datasets.yml* points there and is resolved relative to the
egon-data working directory, like every other `data_bundle_egon_data` path.
The directory is read-only — everything the dataset produces goes to the
database — so there is no `input/` subfolder. The same path is declared in
`HGVCharging.sources` (`DatasetSources(files=...)`) per repo convention, and
`data_bundle` is a dependency of `HGVCharging` so the task cannot run before
the bundle has been downloaded.

### Incidental fixes needed to run this

* `datasets/__init__.py` — tolerate concurrent table creation in
  `Dataset.register()` (`IntegrityError`/`ProgrammingError` race when
  several datasets initialise in parallel).
* `electrical_neighbours.py` — skip tasks when no configured scenario needs
  them, so reGon-only runs do not fail on missing eGon2035 inputs.
* `sanity_checks.py` — skip checks not yet implemented for a given scenario
  rather than raising.
* `model_timeseries.py` (MIV) — scenario-scoped lookups.

## Validation

Validated against `grid.egon_etrago_load(_timeseries)` and the new demand
tables, in a Schleswig-Holstein test run for both scenarios:

| | reGon2037 | reGon2045 |
|---|---|---|
| sites / charging points | 196 / 300 | 196 / 300 |
| charging events | 4,178,887 | 4,896,276 |
| flex models | 79 | 79 |
| highway-day fixed loads | 11 | 11 |

Energy conservation, per aggregated flex model (driving load summed over the
year vs. the charging events it was built from): **mean/median/std = 0.0 %**
across all 79 models in both scenarios.

Site-level vs. event-level energy agrees to **< 0.01 %**. The residual is
deterministic, not noise, and fully accounted for:

* the parking-window timeseries is extended by a mirrored 12 h head / 11 h
  tail so windows can straddle Jan 1 / Dec 31, which **adds** 0.2199 % (depot
  profile);
* year-boundary clipping scales or drops the straddling events, which
  **removes** ~0.2119 pp of that;
* net ≈ +0.008 %, matching the observed per-vehicle-class residual.

Highway-day fixed load reproduces the sites table's `el_demand_day_*`
columns to **±0.0000 %** (profile-based, so no event clipping applies).

A national standalone run reproduces the same figures at full scale (6,092
sites, 140,118,213 events, 2,194 flex models).

## Known limitations

* A small number of depots receive very little demand — 89 of 5,740 (1.6 %)
  below one vehicle per day, concentrated in the Hafen/Bahnhof category
  (22.6 % of that category). Investigated; ~0.37 % are OSM keyword-collision
  false positives. Deliberately left in place rather than adding a pruning
  mechanism.
* ~0.31 % of charging events are clipped to the year boundary. These are
  **dropped** from the lowflex fixed load (rather than concentrating a
  year's arrivals onto hour 0) but left untouched in the flex model's
  driving load, matching MIV's regular-scenario behaviour.
* `lowflex` is currently disabled for the reGon scenarios
  (`create_lowflex_scenario: False`). The reGon scenarios have no lowflex
  path anywhere in the pipeline yet — `LowFlexScenario`'s SQL is hardcoded to
  `eGon2035`, and MIV has no reGon trip data — so enabling only the HGV half
  would produce an incomplete scenario.
* `HeavyDutyTransport` is restricted to `eGon2035`; `eGon100RE` is not
  registered as an `EgonScenario` row in this branch and would hit a foreign
  key error.

## Contributor checklist

- [ ] **Tests pass** — still needs a `tox` run before review.
- [x] **Workflow has run in Test mode** — Schleswig-Holstein test runs
      completed for `reGon2037` and `reGon2045` (figures above).
- [x] **Documentation updated** — `docs/data/mobility_demand.rst` covers the
      heavy-duty section; added `docs/reference/` API stubs for the package
      and all five submodules (`db_classes`, `etrago_integration`,
      `fill_tables`, `scenarios`, `spatial_assignment`) and listed
      `hgv_charging` in the `emobility` toctree, so the new module's API docs
      are actually built. Model rationale lives in the companion `HGV/` repo
      (see *Model documentation* above).
- [x] **Dataset versions updated** — `HGVCharging` is new at `0.0.1`;
      `HeavyDutyTransport` bumped `0.0.5` → `0.0.6` because its active
      scenario list changed (`["eGon2035", "eGon100RE"]` → `["eGon2035"]`).
- [x] **`CHANGELOG.rst` note** — added under *Unreleased → Added*.
- [x] **Added to `AUTHORS.rst`**.


src/egon/data/datasets/emobility/motorized_individual_travel/model_timeseries.py
```diff
@@ -369,7 +375,7 @@ def generate_load_time_series(
         decimal=-1,
     )
 
-    if sources.files["original_data"]["model_timeseries"]["reduce_memory"]:
+    if reduce_memory:
         return reduce_mem_usage(load_time_series_df)
     return load_time_series_df
```
nesnoj, line 397: @MoritzSchloesser 
Merging some branches I stumbled across these changes in the MIT part.
While they seem to look ok I wanted to make sure you are aware of these adjustments.
I couldn't figure out the place you make use of `generate_load_time_series()` and hence had to adjust this function? You adapted it to `_generate_hgv_load_time_series()`, right?

nesnoj:
Another issue: Your dataset expects the files to be located in `data_bundle_egon_data/hgv_charging/<scenario>/`.
However, on SP there 2 ZIP files that need to be extracted manually, is this intended? (it might be described in #1437 but as GH has a lot of timeouts currently I cannot check...)
If not, an auto-extraction might be good like in the eMob MIT part?

MoritzSchloesser:
> Another issue: Your dataset expects the files to be located in `data_bundle_egon_data/hgv_charging/<scenario>/`. However, on SP there 2 ZIP files that need to be extracted manually, is this intended? (it might be described in #1437 but as GH has a lot of timeouts currently I cannot check...) If not, an auto-extraction might be good like in the eMob MIT part?

Yes, it was simply because the files were to big to be uploaded to SP. However, an auto-extract sounds reasonable and I'll add it.

src/egon/data/datasets/emobility/motorized_individual_travel/model_timeseries.py
MoritzSchloesser: Yes, seems like I didn't revert these changes after adding more features to the function for hgv and therefore creating a separate function. Still, I would keep the correction of the step size. This allowd for different step sizes and doesn't assume it to always be 15min. 

I'll remove the memory reduction.

src/egon/data/datasets/emobility/hgv_charging/db_classes.py
```diff
@@ -0,0 +1,142 @@
...
+
+from egon.data.datasets.scenario_parameters import EgonScenario
+
+Base = declarative_base()
+
+
+class EgonHgvChargingSite(Base):
+    """demand.egon_hgv_charging_site — one row per physical site per scenario."""
+
+    __tablename__ = "egon_hgv_charging_site"
```
nesnoj: Proposal: `egon_ev_hgv_charging_site`

src/egon/data/datasets/emobility/hgv_charging/db_classes.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+class EgonHgvChargingPoint(Base):
+    """demand.egon_hgv_charging_point — one row per (site × vehicle_class × time_of_day) per scenario.
+
+    cp_id maps 1:1 to one eDisGo charging_park_id and one edisgo_id in the MV topology.
+    num_cp is the number of physical chargers in the group (metadata).
+    """
+
+    __tablename__ = "egon_hgv_charging_point"
```
nesnoj: Proposal: `egon_ev_hgv_charging_point`

src/egon/data/datasets/emobility/hgv_charging/db_classes.py
```diff
@@ -0,0 +1,142 @@
...
+    """demand.egon_hgv_charging_event — individual parking/charging/vacant
+    events, one row per physical slot per event (see docs/adr/0002).
+
+    Column names mirror MIV egon_ev_trip. HGV-specific extras: cp_id, site_id,
+    vehicle_class, bat_cap, slot_id. Columns computed internally by
+    generate_load_time_series (charge_end, last_timestep, flex_*) are NOT
+    stored here.
+    """
+
+    __tablename__ = "egon_hgv_charging_event"
```
nesnoj: Proposal: `egon_ev_hgv_charging_event`

src/egon/data/datasets/emobility/hgv_charging/db_classes.py
```diff
@@ -0,0 +1,142 @@
...
+    """demand.egon_hgv_profile — normalized 8760-h demand shapes per sector.
+
+    All profiles sum to 1 independently (depot, highway day, highway night).
+    Multiply profile[t] * annual_consumption_mwh to get MW at each 1-h timestep
+    (since profile sums to 1 and each step = 1 h, the product is MWh/h = MW).
+    For highway CPs, annual_consumption_mwh is el_demand_day_mwh or el_demand_night_mwh
+    from egon_hgv_charging_site.
+    """
+
+    __tablename__ = "egon_hgv_profile"
```
nesnoj: Proposal: `egon_ev_hgv_profile`

nesnoj (commented):
Proposal for table names for sake of similarity to the M1+N1 tables
(DB model names need to be adjusted accordingly..)

MoritzSchloesser:
### The `park_start == 0` guard is on the wrong code path

`_build_bus_timeseries_events` drops year-boundary-clipped events from the lowflex fixed-load
sum:

```python
# etrago_integration.py:210
real_events = real_events[real_events["park_start"] != 0]
```

The rationale in `DOCUMENTATION.md` §14 is that such an event has "no valid place to be shown
at full power without concentrating a country-wide year's worth of arrivals onto one
timestep." That failure mode is real — but it belongs to `driving_load`, not to the
event-summed load this filter guards.

The two arrays in `_generate_hgv_load_time_series` accumulate differently:

```python
arrays["load"][start:end] += cap                     # rated charger power, per hour
arrays["load"][last_ts]   += last_ts_cap             # fractional remainder
...
arrays["driving_load"][start] += consumption / TIMESTEP_HOURS   # whole energy, one timestep
```

`driving_load` divides each event's entire energy by one hour, so it is unbounded and clipped
events do produce a spike. `load` contributes only `cap` per event, so it is bounded by
installed charger capacity — 6,555 simultaneous arrivals can never sum to more than those
chargers can physically draw.

Measured on the SH run (`reGon2037`, `demand.egon_hgv_charging_event`):

| quantity | MW |
|---|---|
| hour 0 under dumb charging, if clipped events were kept | 380.0 |
| genuine annual peak simultaneous power, hours 1–8759 | 388.3 |
| genuine mean | 123.9 |
| hour 0 in `driving_load` (clipped events kept, as designed) | 1757.2 |

Hour 0 would land at 98% of a normal peak hour — indistinguishable from a busy evening, not a
spike. For comparison the `driving_load` figure is ~1200x its own mean, which is the artifact
the documentation describes.

Confirmed in `grid.egon_etrago_load_timeseries` for the current run, split by bus carrier:

| rows | bus carrier | hour 0 | hour 1 |
|---|---|---|---|
| 11 | `AC` (fixed load, profile-based highway-day) | 0.00 MW | 0.00 MW |
| 79 | `HGV_charger` (flex `driving_load`) | 1757.17 MW | 26.88 MW |

The 1757.17 MW matches `sum(charging_demand) where park_start = 0` = 1,757,165 kWh exactly, so
the mechanism is confirmed rather than inferred.

### Consequences of the guard as placed

1. **Unnecessary energy loss.** ~0.31% of charging energy is dropped from the lowflex fixed
   load to prevent a spike that cannot occur on that path.
2. **lowflex and flex no longer represent the same fleet.** `driving_load` keeps the clipped
   events, the fixed load drops them, so total HGV energy differs between the two scenarios by
   that ~0.31%. Any cross-scenario energy comparison shows a gap that is not a modelling
   result.
3. **The filter also removes genuine midnight arrivals.** `park_start != 0` cannot distinguish
   "clipped from 31 Dec" from "actually arrived at 00:00 on 1 Jan". Arrivals cluster around
   19:00 (`ARRIVAL_MEAN_H`) so the collateral loss is small, but it is collateral.
4. **The exploration notebook and the pipeline disagree.**
   `HGV/notebooks/explore_stage_b copy.ipynb` calls `_data_preprocessing_hgv` /
   `_generate_hgv_load_time_series` directly, bypassing the guard, so its recomputed charging
   load is the unfiltered series and does not match what the pipeline writes for lowflex.

### Suggested fix

Drop the filter from `_build_bus_timeseries_events` and keep clipped events in the
event-summed load:

```python
real_events = events[events["location"] == "charging"].copy()
# no park_start filter -- see below
```

Neither variant is physically exact. Those vehicles really arrived before the simulated year,
so at hour 0 they would already be partway through charging: keeping them overstates hour 0 and
the following few hours slightly, dropping them understates annual energy by ~0.31%. Keeping
them is the better trade — energy is conserved, the error is bounded at ~2% of one peak hour,
and lowflex stays consistent with flex.

If the hour-0 overstatement is judged worth correcting properly, the physical fix is not a
filter but a partial event: give clipped events a reduced remaining `charging_demand` reflecting
the charging they already completed before the year began. That needs a change in
`generate_charging_events` (Stage A), not in `etrago_integration.py`.

`DOCUMENTATION.md` §14 needs updating either way, since it currently justifies the filter with
the `driving_load` failure mode on a path where that mode cannot occur.

### Worth re-checking on the Germany run

380 MW vs a 388 MW peak is close enough that a different fleet composition or depot
distribution could invert it. The check is cheap:

```sql
-- hour 0 under dumb charging, if clipped events were kept
SELECT scenario, round(sum(charging_capacity_grid)::numeric / 1e3, 2) AS hour0_mw
FROM demand.egon_hgv_charging_event
WHERE location = 'charging' AND park_start = 0
GROUP BY 1;
```

Compare against the genuine annual peak; if hour 0 materially exceeds it, the filter earns its
place on this path after all.

nesnoj (approved):
LGTM and results were tested already.
Thx @MoritzSchloesser !

PR #1474, opened 2026-08-07
Features/#1449 adapt dataset heat supply to new scenarios

### LINKED ISSUE #1449: [FEATURE] adapt dataset Heat_Supply to new scenarios
As part of the new version of egon-data each task of the taskgroup “Heat_Supply ” should be modified to include the scenarios reGon2037, reGon2045 and status2024 (updated version of status2023 which also replaces status2019) while discarding the scenarios eGon100RE and status2019 and keeping eGon2035.

This changes must also include all required to follow the new design concept where all tasks are suppose to be executed for all required scenarios (defined in egon-data.configuration.yaml –scenarios) and decide inside the task if something should be executed.

CarlosEpia:
Fixes #1449.

The changes in this branch were already tested for DE and so far the results look ok.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -50,11 +50,16 @@ Changed
   2030/2040 data points, and remove obsolete
   status2019/status2023/eGon100RE handling
   `#1438 <https://github.com/openego/eGon-data/issues/1438>`_
-* Adapt combined_heat_and_power to reGon-scenarios: different handling of 
+* Adapt combined_heat_and_power to reGon-scenarios: different handling of
   eGon-/reGon-scenarios due to the Kraftwerksliste from the NEP now
   containing MaStR-IDs to match MaStR- with NEP-data. Remove all status-
   quo (except status-quo2024) and eGon100RE mentions.
   '#1447 <https://github.com/openego/eGon-data/issues/1447>'_
+* Adapt Heat_Supply TaskGroup to new scenarios: extend the eGon2035
+  NEP-based cascade (heat pump capacity target + remainder gas boilers)
+  to reGon2037 and reGon2045, make PV-rooftop-based building weighting
+  scenario-aware, and remove obsolete eGon100RE/pypsa-eur-sec handling
+  `#1449 <https://github.com/openego/eGon-data/issues/1449>`_
 
 Bug Fixes
 ---------
--- a/docs/data/heat_demand.rst
+++ b/docs/data/heat_demand.rst
@@ -59,7 +59,7 @@ The share of each CTS building of the corresponding HV-MV substation's heat prof
 for each scenario written to the database table
 :py:class:`EgonCtsHeatDemandBuildingShare<egon.data.datasets.electricity_demand_timeseries.cts_buildings.EgonCtsHeatDemandBuildingShare>`.
 The peak heat demand per building, including residential and CTS demand, is calculated in the dataset
-:py:class:`HeatPumps2035 <egon.data.datasets.heat_supply.HeatPumps2035>`
+:py:class:`HeatPumpsCascade <egon.data.datasets.heat_supply.individual_heating.HeatPumpsCascade>`
 and written to table
 :py:class:`demand.egon_building_heat_peak_loads <egon.data.datasets.heat_supply.BuildingHeatPeakLoads>`.
 
--- a/docs/data/heat_demand.rst
+++ b/docs/data/heat_demand.rst
@@ -56,13 +56,11 @@ The disaggregation of the annual CTS heat demand per census cell to buildings is
 done analogous to the disaggregation of the electricity demand, which is in detail
 described in section :ref:`disagg-cts-elec-ref`.
 The share of each CTS building of the corresponding HV-MV substation's heat profile is
-for both the eGon2035 and eGon100RE scenario written to the database table
+for each scenario written to the database table
 :py:class:`EgonCtsHeatDemandBuildingShare<egon.data.datasets.electricity_demand_timeseries.cts_buildings.EgonCtsHeatDemandBuildingShare>`.
-The peak heat demand per building, including residential and CTS demand, in the two
-scenarios eGon2035 and eGon100RE is calculated in the datasets
-:py:class:`HeatPumps2035 <egon.data.datasets.heat_supply.HeatPumps2035>` and
-:py:class:`HeatPumpsPypsaEurSec <egon.data.datasets.heat_supply.HeatPumpsPypsaEurSec>`,
-respectively, and written to table
+The peak heat demand per building, including residential and CTS demand, is calculated in the dataset
+:py:class:`HeatPumps2035 <egon.data.datasets.heat_supply.HeatPumps2035>`
+and written to table
 :py:class:`demand.egon_building_heat_peak_loads <egon.data.datasets.heat_supply.BuildingHeatPeakLoads>`.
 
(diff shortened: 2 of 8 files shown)


MoritzSchloesser:
The new early-return guard in `insert_nep_list_powerplants` covers the case where *no* NEP-covered scenario is configured. There's a closely related case it doesn't catch: when *some but not all* of `eGon2035` / `reGon2037` / `reGon2045` are configured.

With `--scenarios: [status2024, eGon2035]` and `--dataset-boundary: Schleswig-Holstein`, `any(s in scenarios for s in [...])` is `True`, so execution proceeds past the new guard. Only the `eGon2035` branch runs, so `kw_liste_nep21` is populated but `kw_liste_nep25` stays the empty `DataFrame` it was initialised to. The `concat` therefore has no `c2037_capacity` / `c2045_capacity` columns.

Two places then assume all capacity columns are present, so the run fails twice in a row — fixing the first only surfaces the second.

### 1. `insert_nep_list_powerplants` — testmode scaling loop

```python
for col in [
    "capacity",
    "a2035_capacity",
    "b2035_capacity",
    "c2035_capacity",
    "b2040_capacity",
    "c2037_capacity",
    "c2045_capacity",
]:
    kw_liste_nep.loc[
        kw_liste_nep[kw_liste_nep.federal_state.isnull()].index, col
    ] *= population_share()
```

```text
File ".../egon/data/datasets/scenario_capacities.py", line 648, in insert_nep_list_powerplants
    kw_liste_nep.loc[
KeyError: 'c2037_capacity'
```

Restricting the loop to columns that were actually loaded fixes it. Only the `for` statement changes; the loop body is untouched:

```python
# scale all capacity to the respective population share
# Only columns of the NEP lists that were actually loaded are
# present, depending on the configured scenarios
for col in [
    c
    for c in [
        "capacity",
        "a2035_capacity",
        "b2035_capacity",
        "c2035_capacity",
        "b2040_capacity",
        "c2037_capacity",
        "c2045_capacity",
    ]
    if c in kw_liste_nep.columns
]:
    kw_liste_nep.loc[
        kw_liste_nep[kw_liste_nep.federal_state.isnull()].index, col
    ] *= population_share()
```

### 2. `aggr_nep_capacities` — strict column selection

```python
nep_capacities = insert_nep_list_powerplants(export=False)[
    ["federal_state", "scenario", "carrier", "c2035_capacity", "c2037_capacity", "c2045_capacity"]
]
```

```text
File ".../egon/data/datasets/scenario_capacities.py", line 404, in aggr_nep_capacities
    nep_capacities = insert_nep_list_powerplants(export=False)[
KeyError: "['c2037_capacity', 'c2045_capacity'] not in index"
```

Using `reindex` adds the missing columns as `NaN` instead of raising:

```python
# Depending on the configured scenarios, only one of the NEP power plant
# lists may have been loaded, so not all capacity columns are present.
# Missing ones are added as NaN, which yields empty per-scenario
# aggregations below.
nep_capacities = insert_nep_list_powerplants(export=False).reindex(
    columns=[
        "federal_state",
        "scenario",
        "carrier",
        "c2035_capacity",
        "c2037_capacity",
        "c2045_capacity",
    ]
)
```

No fabricated values enter the results: the `reGon` aggregation filters on `nep_capacities["scenario"] == "reGon"`, and with no NEP2025 data there are no such rows, so that branch is empty regardless of the NaN column and `dropna(subset=["nuts"])` clears it.

### Notes

Both conditions have to hold to trigger this: a partial NEP scenario set, **and** `--dataset-boundary != "Everything"` for the first site (full-Germany runs skip the scaling block entirely; the second site fails either way). That's presumably why it hasn't shown up in CI.

This fails on `dev` today, so it's not introduced by this PR — the line numbers above refer to the **current `dev`** (site 1: loop at ~L639, raising statement at ~L648; site 2: ~L404), *before* any of this PR's changes are applied. Since you're already hardening this function against scenario-dependent columns, it may belong here rather than in a separate PR.

With both applied, `insert_data_nep()` completes for `[status2024, eGon2035]` on a Schleswig-Holstein boundary, and `aggr_nep_capacities` returns the expected eGon2035-only rows (`oil`, `others`, `pumped_hydro` for `DEF`) with no spurious `reGon` entries.

Happy to open a separate PR against `dev` instead if you'd rather keep this one scoped.

MoritzSchloesser:
One more thing in `scenario_capacities.py` that this PR touches: `district_heating_input()` is not idempotent, and the new per-scenario loop makes the consequence worse.

It writes via `session.add(...)` / `session.commit()` with no preceding delete, so every execution **appends** another set of rows to `egon_scenario_capacities` (unlike `insert_nep_list_powerplants`, which uses `to_sql(..., if_exists="replace")`). Re-running the `insert-data-nep` task — after a failure further down, or after clearing it in Airflow — therefore duplicates the four `urban_central_*` entries rather than replacing them.

That surfaces as a confusing failure in a *different* dataset. `capacity_per_district_heating_category()` reads those rows indexed by technology:

```python
target_values = db.select_dataframe(
    f"""
    SELECT capacity, split_part(carrier, 'urban_central_', 2) as technology
    FROM {sources.tables['scenario_capacities']}
    WHERE carrier IN (...)
    AND scenario_name = '{scenario}'
    """,
    index_col="technology",
)
```

With duplicates present, `target_values.capacity["heat_pump"]` is a 2-element `Series` instead of a scalar, and the assignment a few lines later fails:

```text
File ".../heat_supply/district_heating.py", line 61, in capacity_per_district_heating_category
    capacity_per_category.loc[
ValueError: cannot reindex on an axis with duplicate labels
```

Observed on a Schleswig-Holstein run with `--scenarios: [status2024, eGon2035]`:

```
 scenario_name |                carrier                | count |  sum
---------------+---------------------------------------+-------+--------
 eGon2035      | urban_central_geo_thermal             |     2 |  79.26
 eGon2035      | urban_central_heat_pump               |     2 |  81.55
 eGon2035      | urban_central_resistive_heater        |     2 |  83.54
 eGon2035      | urban_central_solar_thermal_collector |     2 | 477.13
```

Each pair holds identical values, so the capacities are simply doubled (e.g. 238.563 MW solar thermal counted twice). Nothing warns about it — `HeatSupply` just fails downstream with a pandas indexing error that gives no hint the real cause is duplicated input rows.

The recommendation: delete the scenario's rows before inserting, mirroring what `individual_heating()` already does in `heat_supply/__init__.py`. Inside the new `for scenario in scenarios:` loop, before the `session.add(...)` calls:

```python
db.execute_sql(
    f"""
    DELETE FROM {targets.tables['scenario_capacities']}
    WHERE scenario_name = '{scenario}'
    AND carrier IN (
        'urban_central_heat_pump',
        'urban_central_resistive_heater',
        'urban_central_geo_thermal',
        'urban_central_solar_thermal_collector')
    """
)
```

Scoping the delete to these four carriers matters: `insert_capacities_per_federal_state_nep()` writes other rows for the same `scenario_name`, so an unscoped `DELETE WHERE scenario_name = ...` would remove those too depending on task ordering.

This is pre-existing on `dev` rather than introduced here, but it's worth folding in because the new loop multiplies the exposure — with `[eGon2035, reGon2037, reGon2045]` configured, a single re-run now appends three scenarios' worth of duplicates instead of one. It's also cheap to get wrong from the outside: the failure appears in `HeatSupply`, not in `ScenarioCapacities`, so the connection back to this function is easy to miss.

MoritzSchloesser (approved):

PR #1476, opened 2026-08-10
Fixes/#1435 automated testing fails on all new branches

### LINKED ISSUE #1435: [BUG] Automated testing on fails on all new branches (GitHub Workflow)
**Describe the bug**  

Is this issue with the CI already kown? The pipeline fails for and it states that the DB is not availabale. Below claude inpected the issue and found a possible cause:

Error from Tox setp:
```
  .tox/py310-nocov-linux/lib/python3.10/site-packages/psycopg2/__init__.py:122: in connect
      conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
  E   psycopg2.OperationalError: connection to server at "127.0.0.1", port 59734 failed: Connection refused
  E   	Is the server running on that host and accepting TCP/IP connections?
```

**Proposed Solution**

Result of a quick claude analysis

Diagnosis
It's not a flaky/unavailable DB — it's a code regression that makes DAG construction require a live database. And it's on dev, so it breaks CI for the whole repo, not just your feature/1414 branch.

Root cause: The recently merged feature "Store datasets sources and targets in separate table" ([36e423891](vscode-webview://0psl8knmvlf6i2d9j5ie8joqtegmfdnujmi6h6sl1qshkpr339fi/src/egon/data/datasets/__init__.py)) added a self.register() call to Dataset.__post_init__. So now, constructing any Dataset object immediately opens a PostgreSQL connection:

# src/egon/data/datasets/__init__.py:453
```python
def register(self):
    SourcesTargetsModel.__table__.create(bind=db.engine(), checkfirst=True)  # <-- live connection
    with db.session_scope() as session:
        ...
```
The two failing tests never needed a DB before:

[test_pipeline_importability](vscode-webview://0psl8knmvlf6i2d9j5ie8joqtegmfdnujmi6h6sl1qshkpr339fi/tests/test_egon-data.py) just imports the pipeline — but importing builds every Dataset, which now hits the DB.
[test_uniqueness_of_automatically_generated_final_dataset_task](vscode-webview://0psl8knmvlf6i2d9j5ie8joqtegmfdnujmi6h6sl1qshkpr339fi/tests/test_dataset_class.py#L41) just instantiates two Datasets.
The CI workflow ([build.yml](vscode-webview://0psl8knmvlf6i2d9j5ie8joqtegmfdnujmi6h6sl1qshkpr339fi/.github/workflows/build.yml)) has no Postgres service, so db.engine() connects to 127.0.0.1:<random port> and gets Connection refused. (test_airflow, which really needs a DB, is explicitly skipped — confirming CI was never meant to have Postgres.)

There's a deeper smell too: Airflow re-parses DAG files constantly, so requiring a DB connection just to import the DAG is fragile even in production.

Recommended fix
Make register() tolerant of a missing DB — skip registration with a warning instead of crashing. Registration still happens normally during real pipeline runs (where the DB exists); imports, DAG parsing, and unit tests stop depending on it.

``` python
def register(self):
    try:
        SourcesTargetsModel.__table__.create(bind=db.engine(), checkfirst=True)
        with db.session_scope() as session:
            ...  # existing body
    except OperationalError as e:
        logger.warning(
            f"Could not register sources/targets for '{self.name}' "
            f"(database unavailable): {e}. Skipping registration."
        )
```

This is minimal and low-risk. The alternative — adding a postgres service container to build.yml — is more invasive, slower, and the test suite clearly wasn't designed to assume one.

Want me to apply the tolerant-register() fix? I'd do it on your current branch (feature/1414-add-rail-transport-dataset), though note the underlying bug lives on dev, so ideally the fix lands there too.

**Environment:**
- OS: [e.g. Windows 11, Ubuntu 22.04]
- Python version: [e.g. Python 3.10, Python 3.11]
- Scenario: [e.g. eGon2035, eGon100RE]
- Dataset Boundary: [Everything or Schleswig-Holstein]

CarlosEpia:
Fixes #1435.
## Problem

Automated tests (`Tests, code style & coverage / Python 3.10`) have been failing on **every** branch for months, blocking merges repo-wide. Two tests fail with:

**Root cause:** commit `36e42389` ("Store datasets sources and targets in separate table", 2026-03-12) made `Dataset.__post_init__` unconditionally call `self.register()`, which opens a live database connection (`db.engine()`) to persist the dataset's sources/targets.

Constructing a `Dataset` happens automatically just by **importing** the pipeline module, not by running it. So every test that imports the pipeline (or builds a `Dataset` at all) now requires a database connection, even though it never used to. Our CI never provisions a database on purpose (there's already an existing test, `test_airflow`, explicitly `skipif`'d for exactly that reason), so any code path that now depends on a DB during import breaks CI unconditionally.

## Solution

Wrapped the database work inside `Dataset.register()` (`src/egon/data/datasets/__init__.py`) in a `try/except OperationalError`. If the DB is reachable, behavior is unchanged. Sources/targets get registered exactly as before. If it's not, we log a warning and skip registration instead of crashing.

This doesn't weaken the tests since the two previously-failing tests don't actually test `register()`'s database logic at all:

- `test_pipeline_importability` only asserts that importing the pipeline module doesn't raise. It was never exercising `register()` on purpose. It just happened to explode as a side effect.
- `test_uniqueness_of_automatically_generated_final_dataset_task` only checks that generated task IDs are unique; it doesn't touch `register()`'s insert/update logic either.

So both tests keep verifying exactly what they were designed to verify. They just stop failing due to an unrelated side effect. Nothing about their actual assertions changed.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -75,6 +75,11 @@ Bug Fixes
 * Fix dynamic line rating only being calculated for a single scenario
   due to a stray early return inside the per-scenario loop
   `#1424 <https://github.com/openego/eGon-data/issues/1424>`_
+* Fix automated tests failing on all branches because constructing a
+  Dataset unconditionally opened a database connection to register its
+  sources/targets; registration is now skipped with a warning when no
+  database is available
+  `#1435 <https://github.com/openego/eGon-data/issues/1435>`_
 
 Version 2.0.0 (2025-08-20)
 ==========================
--- a/src/egon/data/airflow/dags/pipeline.py
+++ b/src/egon/data/airflow/dags/pipeline.py
@@ -87,7 +87,10 @@ from egon.data.datasets.pypsaeur import PreparePypsaEur, RunPypsaEur
 from egon.data.datasets.re_potential_areas import re_potential_area_setup
 from egon.data.datasets.renewable_feedin import RenewableFeedin
 from egon.data.datasets.saltcavern import SaltcavernData
-from egon.data.datasets.sanity_checks import SanityChecks
+
+# SanityChecks import is unused while it's excluded from the pipeline,
+# see the comment near its former TaskGroup below.
+# from egon.data.datasets.sanity_checks import SanityChecks
 from egon.data.datasets.scenario_capacities import ScenarioCapacities
 from egon.data.datasets.scenario_parameters import ScenarioParameters
 from egon.data.datasets.storages import Storages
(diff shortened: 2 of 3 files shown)


nesnoj (approved):
Looks good, thx!
The diagnosis is spot on and this unblocks CI, which matters more right now than the shape of the fix.

One thought for a future follow-up though:
`except OperationalError` is broader than "no database". psycopg2 maps a whole class of operational failures onto it: e.g. too many clients already, auth failures, database does not exist, disk-full. Those all become a single warning line and a silent skip.

PR #1485, opened 2026-08-13
Features/#1483 adapt e mobility mit to re gon scenarios

nesnoj:
Fixes #1483.
**Note**: #1488 should be merged first to avoid etrago id conflicts!

Adapts the eMobility MIT dataset to `status2024`, `reGon2037` and `reGon2045`.
Previously it only knew `status2019`/`status2023`/`eGon2035`/`eGon100RE` and
crashed on the current default scenario set (`KeyError` in `extract_trip_file`).

**Four commits, reviewable in order:**

1. `Update EV numbers for MIT scenarios` — NEP 2025 V.2 Scenario C counts
   (34.1 M / 40.6 M) and KBA stock 01.01.2025 for `status2024`.
2. `Fix duplicate land_transport_EV load` — standalone bug fix, see below.
3. `Adapt eMobility MIT to reGon2037, reGon2045 and status2024` — the migration.
4. `Bump dataset version, update changelog and docs`.

**Bug fix (pre-existing, affects `dev`):** `write_model_data_to_db()` calls
`write_to_db()` a second time for the lowflex variant. For dumb charging
scenarios both calls took the same branch and wrote the load under the
scenario's own name, so every MV grid district got two identical
`land_transport_EV` loads — doubled transport demand. This was live for
`status2019`/`status2023` and would have hit `status2024`. Commit 2 is
self-contained and can be cherry-picked.

**Design decisions:**

- *Trip data:* no simBEV runs exist for 2037/2045, so existing runs are reused
  by horizon (eGon2035 run → `status2024`/`reGon2037`, eGon100RE run →
  `reGon2045`). Same precedent as `status2019`/`status2023`. Profiles are drawn
  with replacement, so larger fleets just resample the pool more often — but the
  battery/charging assumptions in `EgonEvMetadata` are those of the reused run,
  not of the target year.
- *Flexible vs dumb charging* is derived from the scenario name via a new
  `is_flexible()` helper, replacing a hard-coded scenario list — same idiom as
  `scenario_capacities.py` and `chp/`.
- *Remaining-MVGD tasks:* the four `generate_model_data_*_remaining` functions
  are replaced by one generic function whose task is generated per configured
  scenario, so grid districts beyond the parallel bunches can no longer be
  dropped silently when a scenario is added.
- Obsolete `status2019`, `status2023` and `eGon100RE` entries removed.

### 📋 Pull Request Guidelines

[...]

PR #1287, opened 2025-06-19
Features/#1284 mitigate next etrago id issues

ClaraBuettner:
Fixes #1284 
Fixes #1364

## Before merging into `dev`-branch, please make sure that

- [x] the `CHANGELOG.rst` was updated.
- [x] new and adjusted code is formated using `black` and `isort`.
- [x] the `Dataset`-version is updated when existing datasets are adjusted.
- [x] the workflow is running successful in `test mode`.
- [ ] the workflow is running successful in `Everything` mode.

PR #1488, opened 2026-08-18
Fixes/#1487 next etrago id issues

### LINKED ISSUE #1487: DB `UniqueViolation`: `next_etrago_id()` mitigation is not working properly
Found while executing #1485 - some emobility MIT bunch tasks failed raising a DB `UniqueViolation`.

@CarlosEpia in this case `h2_neighbours_egon2035()` reserved only one `bus_id` instead of 14 which are committed to the DB. At the same time task `generate_model_data_eGon2035_bunch1800-215()` tried to write to the DB, that's why it failed.

**Root cause:** In #1287 colliding etrago IDs were mitigated, I was not sure if the implementation will work for DE in those days, see comment history. Apparently, some parts were not implemented thoroughly: I analyzed the parallel tasks and the id obtainment in `h2_neighbours_egon2035()` reserves one id from `grid.etrago_bus_id_seq` but writes N buses with consecutive ids starting there. The sequence advances by 1; the table gains N occupied ids. So the window X+1 … X+N-1 is simultaneously reported free by the sequence and occupied in the table. Any task drawing from that window gets a duplicate key.

Code of this example:
https://github.com/openego/eGon-data/blob/9b4162d78ce93eae6b25050e6eb22095ee7e9c1d/src/egon/data/datasets/pypsaeur/__init__.py#L130-L132

Instead, the count should be passed to `next_etrago_id()` like
```
h2_bus["bus_id"] = db.next_etrago_id("bus", len(h2_bus.index)
```

**Moreover, I found multiple other occurences where IDs are not reserved correctly.**
I'll prepare a fix for all of them.

nesnoj:
Fixes #1487.
**Note:** This issue should be merged before #1485 to avoid etrago id conflicts!

Follow-up to #1287, whose mitigation was not applied thoroughly. Every ID
written to a `grid.egon_etrago_*` table must be reserved from the matching
sequence; three places did not do that, so the sequences drifted behind the
IDs actually present and concurrent inserts failed with `UniqueViolation`.

**Fixes**

- `pypsaeur.h2_neighbours_egon2035()` reserved one `bus_id` and derived a
  range of N from it. Now passes the count. This is the one that broke the
  MIT bunch tasks in #1485.
- `electrical_neighbours.insert_storage_units_sq()` drew `storage_id` for
  foreign pumped hydro from the `store` sequence instead of `storage`, so
  the `storage` sequence never advanced for those rows. Its sibling block
  for batteries 70 lines below already did it correctly.
- `DSM_cts_ind.aggregate_components()` asked for the `"Link"`/`"Store"`
  sequences. This only worked because PostgreSQL case-folds unquoted
  identifiers.

**`next_etrago_id()` hardening**

- Lower-cases the component name, so lookups no longer rely on that folding.
- Returns a list whenever a count is given, instead of a scalar for
  `count == 1`. Eight call sites assign the result to a DataFrame index and
  would have crashed with a `TypeError` on a single-row frame.
- Documents that IDs must be reserved one call per ID.

Dataset versions bumped: `SolvePypsaEur` 0.0.43 → 0.0.44,
`ElectricalNeighbours` 0.0.15 → 0.0.16. `DsmPotential` is unchanged, the
case fix alters no behaviour.

**Checked, no change needed:** all ~100 `next_etrago_id()` call sites were
reviewed. `motorized_individual_travel` is clean — it reserves exactly one
ID per row from the correct sequence in all four of its writes; it was the
victim here, not a cause, simply because it makes ~4 sequence draws per MV
grid district per scenario.

@CarlosEpia **Before re-running on an existing database,** the already desynchronised
sequences need realigning, otherwise the poisoned ID blocks remain:

```sql
SELECT setval('grid.etrago_bus_id_seq',     (SELECT MAX(bus_id)     FROM grid.egon_etrago_bus));
SELECT setval('grid.etrago_storage_id_seq', (SELECT MAX(storage_id) FROM grid.egon_etrago_storage));
```

## 🧑‍💻 Contributor Checklist

- [x] Workflow has run at least once in **Test mode**

---

## 🔍 Reviewer Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -80,6 +80,14 @@ Bug Fixes
   sources/targets; registration is now skipped with a warning when no
   database is available
   `#1435 <https://github.com/openego/eGon-data/issues/1435>`_
+* Fix eTraGo ID sequences getting out of sync with the IDs actually
+  written, which made concurrent inserts fail with ``duplicate key value
+  violates unique constraint``: ``h2_neighbours_egon2035`` derived a range
+  of IDs from a single reserved one, and foreign pumped hydro storages
+  drew their ``storage_id`` from the ``store`` sequence. ``next_etrago_id``
+  now ignores the case of the component name and always returns a list
+  when a count is given
+  `#1487 <https://github.com/openego/eGon-data/issues/1487>`_
 
 Version 2.0.0 (2025-08-20)
 ==========================
--- a/src/egon/data/datasets/DSM_cts_ind.py
+++ b/src/egon/data/datasets/DSM_cts_ind.py
@@ -1061,13 +1061,13 @@ def aggregate_components(df_dsm_buses, df_dsm_links, df_dsm_stores):
     df_dsm_stores["bus"] = bus_id
 
     # select new link_ids for aggregated links
-    link_id = db.next_etrago_id("Link", len(df_dsm_links.index))
+    link_id = db.next_etrago_id("link", len(df_dsm_links.index))
 
     df_dsm_links["link_id"] = link_id
 
     # select new store_ids to aggregated stores
 
-    store_id = db.next_etrago_id("Store", len(df_dsm_stores.index))
+    store_id = db.next_etrago_id("store", len(df_dsm_stores.index))
 
     df_dsm_stores["store_id"] = store_id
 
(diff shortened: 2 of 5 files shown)


nesnoj:
Hey @CarlosEpia , may I ask you to review this one?
If that'd be okay, please make sure you reset the ids (see note above) if you want to apply it to the current DE run.

It ran successfully in SH.

CarlosEpia (approved):
I went through the code changes and they look correct. If you already tested for SH that would be enough for me to merge this changes.

Issue #1475, opened 2026-08-07
[FEATURE]  Rework status2024 demand data: rerun generation independently instead of proxying eGon2035, and verify HP capacity placeholder


MoritzSchloesser:
## Background

`status2024` (the status-quo scenario) currently gets its demand data (`egon_peta_heat`, CTS heat demand share/profile, DH-exclusion mapping) by proxying live reads to `eGon2035` at query time, via `demand_source_scenario()`/`pv_source_scenario()` mappings configured in `datasets.yml`. PV rooftop weighting reads the pre-existing `status_quo` rows directly (unaffected by this issue).

This was an intentional interim step — see ADR-0002 (`docs/adr/0002-rerun-demand-generation-for-status2024-instead-of-proxying.md`) — chosen because getting the HP distribution running took priority over building the "real" data path. It's now superseded by a design decision made once the broader multi-scenario pipeline shape (`status2024`, `eGon2035`, `reGon2037`, `reGon2045`, each task iterating all configured scenarios) became clear.

## What needs to change

1. **Replace the demand proxy with an independent rerun.**

   Rather than `status2024` reading `eGon2035`'s rows live via `demand_source_scenario()`, rerun the demand-generation tasks a second time with `status2024` as their own target scenario, producing independent output rows in `egon_peta_heat`, `egon_cts_heat_demand_building_share`, `egon_etrago_heat_cts`, `egon_map_zensus_district_heating_areas`.

   - Rerun was chosen over copying `eGon2035`'s finished rows so that status2024-specific input data can be substituted later without unwinding a separate copy step.
   - Known consequence to watch for: `status2024` and `eGon2035` demand data will only match because the generation *logic* is currently identical for both — any scenario-dependent behavior already latent in that logic (e.g. weather year, building stock assumptions) will surface as a real divergence once this changes, rather than being silently proxied over. Treat that as a signal to investigate, not a bug in the rework itself.
   - Once this lands, `demand_source_scenario()` / `pv_source_scenario()` (`scenario_parameters/__init__.py`) and the `heat_pumps_status_quo` mapping block in `datasets.yml` can likely be removed for the demand side (PV mapping stays).

2. **Verify the placeholder national HP capacity.**

[...]

PR #1495, opened 2026-08-20
Features/1477 fixed floor regon heat pumps

### LINKED ISSUE #1477: [FEATURE] Make future-scenario HP distribution treat status2024's assignments as a fixed floor
## Background

Heat pumps for the `status2024` status-quo scenario are distributed per-building and written to `demand.egon_hp_capacity_buildings` (see the HP-distribution pipeline: `insert_capacities_status_quo` → `cascade_per_technology`/`cascade_heat_supply_indiv` → `determine_buildings_with_hp_in_mv_grid`, all in `heat_supply/individual_heating.py`).

The design intent (unchanged since the original status-quo planning) is that **future scenarios must treat status2024's already-placed heat pumps as a fixed floor**: a building that has a status2024 HP keeps at least that capacity in every later scenario, and only *additional* capacity beyond the status2024 stock gets newly distributed to other buildings.

This is **not yet implemented**. `HeatPumps2035` (`individual_heating.py:321`) and its per-building distribution (`determine_hp_cap_buildings_pvbased_per_mvgd`, `individual_heating.py:1632`, called from `determine_hp_cap_peak_load_mvgd_ts_2035` at `:1955`/`:2007`/`:2145`) currently distribute their scenario's full target capacity from scratch, with no awareness of what status2024 already assigned.

## What needs to change

For each future scenario (`eGon2035`, `reGon2037`, `reGon2045`), the per-building HP distribution needs to:

1. **Read status2024's existing assignments** from `demand.egon_hp_capacity_buildings` (filtered to `scenario = 'status2024'`) for the buildings in the MV grid being processed.
2. **Pre-assign those buildings** their status2024 capacity as a floor, before distributing anything new.
3. **Distribute only the remaining capacity** — `(future scenario's target capacity for the grid) − (status2024 capacity already placed in that grid)` — to the buildings not yet equipped, using the existing PV-weighted logic.

This affects `determine_hp_cap_buildings_pvbased_per_mvgd` (`individual_heating.py:1632`) and its callers for each future scenario.

## Open questions to resolve before implementation

- What happens if a future scenario's target capacity for a grid is *less* than what status2024 already placed there? (Floor logic implies the future scenario can't go below status2024 — does the target ever need to be clamped up, or is this not expected to occur in practice?)
- Does a building keep *exactly* its status2024 capacity, or can it receive *more* than that if the future-scenario distribution logic would otherwise assign it additional capacity?
- Do `reGon2037` and `reGon2045` chain off `eGon2035`'s result (i.e. treat `eGon2035`'s total placed HPs, including its own status2024 floor, as *their* floor), or does each future scenario independently floor only against `status2024` directly?

## Out of scope

- The status2024 HP-distribution pipeline itself (input data, capacity figures) — already implemented and verified; see the status2024 HP implementation work this issue follows on from.
- Any scenario-string handling issues unrelated to the HP floor logic (tracked separately).

## References

- `heat_supply/individual_heating.py`: `HeatPumpsStatusQuo` :233 · `HeatPumps2035` :321 · `determine_hp_cap_buildings_pvbased_per_mvgd` :1632 · `determine_hp_cap_peak_load_mvgd_ts_2035` :1955
- Output table: `demand.egon_hp_capacity_buildings`

MoritzSchloesser:
## Make future-scenario HP distribution treat status2024's assignments as a fixed floor

Closes #1477.

Heat pump assignments were drawn independently in every scenario, so buildings gained and lost
heat pumps at random even though total installed capacity only grows along the timeline.
Measured before this change: only 29% of `status2024`'s heat pump buildings still had one in
`reGon2037`, and 47% of `reGon2037`'s survived into `reGon2045`.

This makes the assignment propagate along a **floor chain**: a building with a heat pump in one
scenario keeps it, at no less than its inherited capacity, in the next.

### Design

Recorded in ADR-0003; the essentials:

**Chain:** `status2024 -> reGon2037 -> reGon2045`. Each scenario inherits its predecessor's
*result*, not `status2024` directly, so the stock is monotonically non-decreasing along the
chain.

**`eGon2035` is deliberately excluded.** It is a 2021-NEP legacy scenario rather than part of
the reGon timeline, and giving it a floor would change already-reviewed `eGon2035` results for
no modelling benefit. It keeps its independent PV-weighted distribution. The consequence is
accepted explicitly: `eGon2035` remains internally incoherent with respect to `reGon2037` in
the way described above.

**Inherited capacity is a lower bound, not a fixed value.** Where a building's peak heat demand
has grown since the inherited scenario, its heat pump is raised to the minimum the sizing rule
requires for the current scenario. This matters in practice, not just in theory: every one of
the 126,832 buildings inherited from `reGon2037` into `reGon2045` has a *higher* peak demand in
`reGon2045` (mean +1.62%), and with capacity frozen 113,777 of them — about 90% — would have
been below the capacity needed to cover their own building, the worst at 73% of it. That would
surface downstream as insufficient heat supply rather than as a conservative approximation.
Capacity is still never reduced, so the ~20-year service life argument is untouched: up-scaling
represents replacing or supplementing a unit its building has outgrown.

**Up-scaling comes out of the distributable budget**, not added to the grid total, so
conservation is unaffected: a grid whose inherited stock grows absorbs more of its own target
and has correspondingly less left for newly equipped buildings.

**The chain is configured, not hardcoded.** `datasets.yml` gains
`demand_timeseries_mvgd.hp_floor_chain`; extending the chain (e.g. `reGon2050`) needs no Python
change. If a run omits a middle link, the floor walks back to the nearest predecessor that has
rows — a weaker but still valid constraint. If no predecessor has any rows the scenario is
distributed unfloored **with a warning**, since running a single future scenario standalone is a
legitimate way to test the distribution itself.

### Implementation

Per MV grid, in `determine_hp_cap_buildings_pvbased_per_mvgd()`:

```
inherited  = predecessor's assignments for this grid's candidate buildings
floored    = max(inherited, sizing-rule minimum for this scenario)   # never reduced
remaining  = hp_cap_grid - sum(floored)
new        = existing PV-weighted selection + desaggregate_hp_capacity(remaining)
```

Floored buildings are exempt from the proportional scaling in `desaggregate_hp_capacity()`; only
the remainder is distributed, using the existing selection unchanged. Both invariants stay
exact: no floored building is undersized for its own scenario, and the distributed total still
equals `hp_cap_grid`.

Task ordering comes free from the existing `Dataset` task-graph API — a `tuple` of task graphs
runs in order, and `HeatPumpsCascade` accumulates its per-scenario task sets into one flat
tuple, so all of `reGon2037`'s parallel bulks complete before any `reGon2045` task starts. The
task tuple is built in the order declared by `hp_floor_chain` rather than in config-list order,
so listing the scenarios the other way round cannot silently invert the chain.

### Validation

Full run, `--dataset-boundary: Schleswig-Holstein`,
`--scenarios: [status2024, eGon2035, reGon2037, reGon2045]`. 27 checks, all passing.

| property | result |
|---|---|
| Conservation, all four scenarios | ≤ 0.003 MW delta, 100.0001% of target |
| Sizing rule, all four scenarios | 0 undersized, min ratio 1.0 |
| Floor containment, both edges | 100.0% of eligible buildings, 0 unexplained |
| Floored capacity reduced | 0 of 14,526 and 0 of 126,832 |
| Up-scaling applied | 113,777 of 126,832 raised, 0 left undersized |
| HP inside a DH area | 0 in every scenario |

Installed capacity per scenario is unchanged by the floor — it redistributes *which* buildings
hold capacity, not how much:

| scenario | buildings | capacity | target |
|---|---|---|---|
| status2024 | 15,069 | 349.42 MW | 349.42 MW |
| eGon2035 | 103,274 | 1654.37 MW | 1654.37 MW |
| reGon2037 | 143,992 | 2410.65 MW | 2410.65 MW |
| reGon2045 | 228,181 | 3403.27 MW | 3403.27 MW |

`reGon2045` holds ~2,340 fewer buildings than before up-scaling at the same total, which is the
intended trade-off: capacity spent raising inherited units is capacity not available for new
ones.

### Not 100% containment against the raw predecessor count — and why that is correct

Containment is measured against **eligible** predecessor buildings, i.e. those still candidates
for individual heating in the successor. District heating expands from 5,091 zensus cells in
`status2024` to 7,645 in `reGon2037` and 16,294 in `reGon2045`, and a building that joins a DH
area is supplied by the network instead — it is never given both. Every drop-out is attributed:

| edge | missing | moved to DH | no peak load | no zensus cell | **unexplained** |
|---|---|---|---|---|---|
| status2024 -> reGon2037 | 543 | 543 | 0 | 0 | **0** |
| reGon2037 -> reGon2045 | 17,160 | 17,160 | 0 | 0 | **0** |

Two details are needed to classify these correctly, and both initially produced phantom "floor
violations" during review: a building's zensus cell can live in any of three mapping tables
(`egon_heat_timeseries_selected_profiles` for the residential path,
`openstreetmap.egon_cts_buildings` for CTS, and
`boundaries.egon_map_zensus_buildings_filtered_all`), and a building can span several cells (up
to four observed), so DH membership must be an "any cell" test.

### Review notebook

`hp_review_all_scenarios.ipynb` (not part of this PR) covers all four scenarios: table
contents, unit and null checks, conservation against target both nationally and per MV grid,
the sizing rule, inherited-capacity behaviour along each chain edge, scenario comparison, the
4x4 building-set overlap matrix with chain-edge attribution, per-grid distributions, PV
weighting, and 300 m windows with all four scenarios overlaid by marker.

### Still to do, not in this PR

- **Status-quo sanity checks.** `SanityChecks` is currently commented out of `pipeline.py` and
  its checks only cover `eGon2035`/`eGon100RE`. The four chain-transfer checks used to validate
  this work are candidates for it.
- Depends on the PR for #1475 (`/16` double-scaling fix, `pv_source_scenario`, PV
  deduplication), which is merged into this branch; the diff against `dev` will shrink to the
  floor alone once that lands.

### Observations for separate issues

- Residential heat demand in `egon_peta_heat` has a large implausible tail — 25% of cells below
  1 MWh/a — producing heat pumps rated in single-digit watts. Upstream of this work; separate
  issue filed.
- 311 CTS buildings have a peak load in `reGon2037` but not in `reGon2045` (311/311 CTS, only
  27 also residential), so they silently lose candidacy in that scenario. Likely the
  `catch_missing_buidings()` path already flagged in the code as a known SH-cutout issue
  referencing #990. Does not affect the floor.
- `demand.egon_hp_capacity_buildings.hp_capacity` is `real` (24-bit mantissa) while capacity
  targets are `double precision`. The per-building truncation accumulates to a ~1e-9 relative
  offset in the stored total (e.g. −0.0023 MW on 2410.65 MW). Physically irrelevant, but it
  means conservation checks against this table need a relative tolerance rather than an
  absolute one.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,8 +21,16 @@ develop-eggs
 lib
 lib64
 venv*/
+.venv*/
 pyvenv*/
 pip-wheel-metadata/
+
+# Local runtime config (contains DB credentials) and scratch/test scripts
+egon-data.configuration.yaml
+scratch/
+
+# Session planning docs (local use only, not project documentation)
+STATUS2024_HP_IMPLEMENTATION_PLAN.md
 
 # Installer logs
 pip-log.txt
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -84,6 +84,19 @@ Changed
 Bug Fixes
 ---------
 
+* Run the heat pump floor chain in order across datasets: HeatPumpsCascade now
+  depends on HeatPumpsStatusQuo, so cascade scenarios can no longer read a
+  status quo that is still being written and silently lose their inherited
+  floor. A partially written predecessor is detected and raises instead of
+  being treated as an empty floor.
+  `#1477 <https://github.com/openego/eGon-data/issues/1477>`_
+* Exclude buildings with zero heat peak load from heat pump allocation, so they
+  no longer end up in egon_hp_capacity_buildings with a capacity of 0.
+  `#1477 <https://github.com/openego/eGon-data/issues/1477>`_
+* Create the heat pump output tables in a single shared task instead of racing
+  checkfirst=True creates from the per-scenario delete tasks, which could fail
+  with a UniqueViolation under the LocalExecutor.
+  `#1477 <https://github.com/openego/eGon-data/issues/1477>`_
 * Fix URL of BASt traffic data
   `#1347 <https://github.com/openego/eGon-data/issues/1347>`_
 * Discard scenario_path tasks
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,8 @@ Added
   (at exactly its inherited capacity) in later scenarios. eGon2035 is not on the
   chain and keeps its independent distribution.
   `#1477 <https://github.com/openego/eGon-data/issues/1477>`_
+* Use MaStR data for home_batteries for allocation for all scenarios (status + future)
+  `#1470 <https://github.com/openego/eGon-data/issues/1470>`_
 
 Changed
 -------
@@ -65,6 +67,19 @@ Changed
   to reGon2037 and reGon2045, make PV-rooftop-based building weighting
   scenario-aware, and remove obsolete eGon100RE/pypsa-eur-sec handling
   `#1449 <https://github.com/openego/eGon-data/issues/1449>`_
+* Unify NEP input data into a single ``NEP`` folder in the data bundle
+  `#1479 <https://github.com/openego/eGon-data/issues/1479>`_
+* Adapt eMobility MIT to the reGon scenarios: add trip, scenario
+  variation and lowflex configuration for status2024, reGon2037 and
+  reGon2045, replace the per-scenario ``generate_model_data_*_remaining``
+  tasks with one generically generated task per configured scenario,
+  derive the flexible/dumb charging distinction from the scenario name
+  instead of a hard-coded scenario list, and remove obsolete
+  status2019/status2023/eGon100RE handling
+  `#1483 <https://github.com/openego/eGon-data/issues/1483>`_
+* Allign Methodology for laoding NEP target values for Battery storage for all scenarios
+  `#1471 <https://github.com/openego/eGon-data/issues/1471>`_
+
 
 Bug Fixes
 ---------
@@ -90,6 +105,9 @@ Bug Fixes
   sources/targets; registration is now skipped with a warning when no
   database is available
   `#1435 <https://github.com/openego/eGon-data/issues/1435>`_
+* Fix ONEP area key lookup so status scenarios work in wind offshore:
+  accept both plain codes and codes with a bracketed description
+  `#1480 <https://github.com/openego/eGon-data/issues/1480>`_
 * Fix eTraGo ID sequences getting out of sync with the IDs actually
   written, which made concurrent inserts fail with ``duplicate key value
   violates unique constraint``: ``h2_neighbours_egon2035`` derived a range
@@ -98,6 +116,28 @@ Bug Fixes
   now ignores the case of the component name and always returns a list
   when a count is given
   `#1487 <https://github.com/openego/eGon-data/issues/1487>`_
+* Fix ``fill_etrago_gen`` crashing with an ``IndexError`` for offshore wind
+  power plants sited at ONEP/NEP connection points west of 5.5° longitude,
+  which fell outside the hardcoded ``Germany-offshore`` ERA5 cutout extent
+  and therefore had no ``wind_offshore`` feed-in timeseries; the cutout and
+  the matching weather-cell filter in ``renewable_feedin.py`` now cover the
+  full extent of the connection points defined in ``map_ONEP_areas()``
+  `#1498 <https://github.com/openego/eGon-data/issues/1498>`_
+* Fix eMobility MIT writing a duplicate ``land_transport_EV`` load for
+  dumb charging scenarios: the lowflex pass took the same branch as the
+  regular pass and inserted a second identical load under the scenario's
+  own name, doubling transport demand in status quo scenarios
+  `#1483 <https://github.com/openego/eGon-data/issues/1483>`_
+* Fix conventional non-CHP and pumped hydro power plant allocation for the
+  reGon2037/reGon2045 scenarios: the NEP2025 Kraftwerksliste has no
+  ``bnetza_id`` column (only ``mastr_id``) and stores the CHP flag as
+  lowercase ``ja``/``nein``, so ``select_nep_power_plants`` /
+  ``select_nep_pumped_hydro`` crashed with ``UndefinedColumn`` and, once
+  that was worked around, allocated nothing because every ``chp = 'Nein'``
+  filter missed. ``bnetza_id`` is now only selected for eGon2035,
+  pumped hydro uses ``mastr_id`` for the reGon path, and the CHP flag is
+  normalized to ``Ja``/``Nein`` on import
+  `#1510 <https://github.com/openego/eGon-data/issues/1510>`_
 
 Version 2.0.0 (2025-08-20)
 ==========================
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,11 @@ Added
   `#1352 <https://github.com/openego/egon-data/issues/1352>`_
 * Add standardized sources and targets definitions across dataset modules
   `#1283 <https://github.com/openego/egon-data/issues/1283>`_
+* Add a fixed floor for heat pumps along the scenario chain
+  status2024 -> reGon2037 -> reGon2045, so a building with a heat pump keeps it
+  (at exactly its inherited capacity) in later scenarios. eGon2035 is not on the
+  chain and keeps its independent distribution.
+  `#1477 <https://github.com/openego/eGon-data/issues/1477>`_
 
 Changed
 -------
--- a/docs/data/mobility_demand.rst
+++ b/docs/data/mobility_demand.rst
@@ -3,12 +3,16 @@
 Motorized individual travel
 ++++++++++++++++++++++++++++
 
-The electricity demand data of motorized individual travel (MIT) for both the eGon2035
-and eGon100RE scenario is set up in the
+The electricity demand data of motorized individual travel (MIT) is set up for all
+configured scenarios in the
 :py:class:`MotorizedIndividualTravel<egon.data.datasets.emobility.motorized_individual_travel.MotorizedIndividualTravel>`
 dataset.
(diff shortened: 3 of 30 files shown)


MoritzSchloesser:
Ran into a race condition in `heat_supply.individual_heating` that's relevant here: `delete_*_status_quo` and `delete_*_2035` task pairs each independently call `<Table>.__table__.create(bind=engine, checkfirst=True)` on the same table, with no dependency edge between the two tasks. Under `LocalExecutor` they can run concurrently, and `checkfirst=True` is a non-atomic check-then-create — both tasks see the table doesn't exist yet and both try to `CREATE TABLE`, so the loser fails.

Observed failure:

```
sqlalchemy.exc.IntegrityError: (psycopg2.errors.UniqueViolation)
duplicate key value violates unique constraint "pg_type_typname_nsp_index"
DETAIL:  Key (typname, typnamespace)=(egon_building_heat_peak_loads, 31409542) already exists.

[SQL:
CREATE TABLE demand.egon_building_heat_peak_loads (
	building_id INTEGER NOT NULL,
	scenario VARCHAR NOT NULL,
	sector VARCHAR NOT NULL,
	peak_load_in_w REAL,
	PRIMARY KEY (building_id, scenario, sector)
)
]
```

`delete-heat-peak-loads-2035-reGon2037` failed this way; its sibling `delete-heat-peak-loads-status-quo-2024` started 18ms earlier in the same run and won the race. The failure cascaded into all 10 downstream `determine-hp-capacity-reGon2037-mvgd-bulk0..9` tasks being marked `upstream_failed`, stalling that scenario's heat-pump-capacity stage until manually cleared.

Same pattern exists for two other pairs in the same file:
- `delete_hp_capacity_status_quo` / `delete_hp_capacity_2035` → `EgonHpCapacityBuildings.__table__.create(checkfirst=True)`
- `delete_mvgd_ts_status_quo` / `delete_mvgd_ts_2035` → `EgonEtragoTimeseriesIndividualHeating.__table__.create(checkfirst=True)`

Two possible fixes: hoist each table's creation into a single shared upstream task both scenario-specific deletes depend on, or catch `IntegrityError`/`UniqueViolation` around the `create()` call and treat "someone else just created it" as success (with a `session.rollback()` first, since the failed statement leaves the connection's transaction in a failed state).

Seen on commit `8dc7727b1482cef0caf4f0850224f102be82a263` (branch `server-run-24-08`), Airflow 2.11.0 / `LocalExecutor`, during a nationwide multi-scenario run.

MoritzSchloesser:
Found an issue while validating the output of this stage: **644 buildings end up with `hp_capacity = 0`** in `egon_hp_capacity_buildings` for this run (329 in `status2024`, 315 in `reGon2037` — the `reGon2037` ones are the same buildings inherited via the floor chain from `status2024`). All 644 have a summed `peak_load_in_w` of exactly `0` in `egon_building_heat_peak_loads`.

Traced it to `determine_buildings_with_hp_in_mv_grid` / `catch_missing_buidings`: a building with zero peak heat demand still gets `min_hp_cap_per_building = 0` (not dropped), and `catch_missing_buidings` only excludes a building if it's **absent** from the peak-load index, not if its value is `0`. So a zero-peak building can still be drawn into `buildings_with_hp` by the PV-weighted `np.random.choice`, and `desaggregate_hp_capacity`'s `min_hp_cap_per_building * fac + min_hp_cap_per_building` correctly evaluates to `0` for it — the math is internally consistent, but the practical result is a "heat pump" building with 0 MW capacity, which is what the module's own `n_nonpositive` sanity check (see e.g. a review notebook checking `hp_capacity <= 0`) is designed to catch.

`catch_missing_buidings`'s docstring already flags this general area as unfinished (`# TODO maybe remove after succesfull DE run`, `# Might be fixed in #990`) — filtering the candidate pool to `peak_heat_demand > 0` before selection (in `catch_missing_buidings` or just before the `np.random.choice` call in `determine_buildings_with_hp_in_mv_grid`) looks like it would close this.

Seen on commit `8dc7727b1482cef0caf4f0850224f102be82a263` (branch `server-run-24-08`), during a nationwide (`--dataset-boundary: Everything`) multi-scenario run.

MoritzSchloesser:
Found a real bug while validating this scenario chain on a nationwide (`--dataset-boundary: Everything`) `status2024`/`reGon2037` run: **`determine-hp-capacity-reGon2037-mvgd-bulk*` tasks are not fully gated on all `determine-hp-capacity-status2024-mvgd-bulk*` tasks completing.**

`status2024-mvgd-bulk1` (MV grids 36660–36675) ran from `2026-09-01 05:35:43` to `13:47:33`. Several `reGon2037` bulk tasks (`bulk2`–`bulk6`) had already started at `07:51:32` — **~6 hours before** `status2024-bulk1` committed its rows for those exact grids. `get_inherited_hp_capacity()` reads the predecessor scenario's rows directly from the DB with no synchronization beyond DAG task ordering, so any `reGon2037` grid processed during that overlap window found no predecessor capacity and lost its floor entirely.

Confirmed end-to-end on one example: building `11549602` had `hp_capacity=0.36084956` MW in `status2024`, but only `0.22782852` MW in `reGon2037` — which matches `peak_load_in_w(reGon2037) * 24/18 / 1.7` to 5 significant figures, i.e. it was resized purely from its own current peak load with zero trace of the inherited floor.

Same 15 MV grids (36659–36675) show up in two places downstream:
- **121 of 264,111 inherited buildings had their capacity *reduced*** along the `status2024 -> reGon2037` edge (should only ever stay the same or increase, per `determine_hp_cap_buildings_pvbased_per_mvgd`'s own docstring).
- **~816 of 818 "unexplained" floor-chain drop-outs** (eligible candidates that lost their heat pump with no attributable DH-move/no-peak-load/no-zensus-map cause) live in the same 15 grids.

Suggested fix: an explicit Airflow dependency from all `status2024`-scenario `determine-hp-capacity` bulk tasks to all `reGon2037`-scenario ones (or a single barrier task in between), rather than whatever partial/implicit ordering exists now.

PR #1505, opened 2026-08-28
Fix/#1489 doubling of pv rooftop capacity for statusquo

### LINKED ISSUE #1489: [BUG] pv_rooftop_to_buildings writes status-quo scenario rows twice, doubling PV capacity
## Summary

`pv_rooftop_to_buildings()` writes every status-quo-scenario row to
`supply.egon_power_plants_pv_roof_building` **twice**. The rows are exact duplicates, down to
the same `gens_id`, so the scenario's PV capacity is exactly 2x overstated.

Observed on a Schleswig-Holstein run with `--scenarios: [status2024, eGon2035]`:

```
  scenario  |  rows  | distinct_b | sum_cap  | sum_cap_dedup
------------+--------+------------+----------+---------------
 eGon2035   | 222414 |     222414 | 2733.567 |      2733.567
 status2024 | 233700 |     116850 | 3722.069 |      1861.035
```

`eGon2035` is unaffected. Only status-quo scenarios are duplicated.

Example of a duplicated pair — identical in every column including `gens_id`:

```
 index  |  scenario  | bus_id | building_id |     gens_id     | capacity | voltage_level
--------+------------+--------+-------------+-----------------+----------+---------------
 158674 | status2024 |   1666 |      985847 | SEE907528493492 |   0.1083 |             7
  41824 | status2024 |   1666 |      985847 | SEE907528493492 |   0.1083 |             7
```

## Cause

In `pv_rooftop_buildings.py`, the accumulator `all_buildings_gdf` is **seeded with the
status-quo rows before the scenario loop starts** (`:2430`):

```python
all_buildings_gdf = (
    desagg_mastr_gdf.assign(scenario=status_quo)   # status2024 rows already in
    .reset_index()
    .rename(columns={"geometry": "geom"})
    .set_geometry("geom")
)

scenario_buildings_gdf_sq = all_buildings_gdf.copy()
```

The loop then iterates over `SCENARIOS`, which **includes** `status2024`, and appends each
scenario's rows to that same accumulator (`:2440`, `:2473`):

```python
for scenario in SCENARIOS:
    if scenario == status_quo:
        scenario_buildings_gdf = scenario_buildings_gdf_sq.copy()   # the same rows again
    ...
    all_buildings_gdf = gpd.GeoDataFrame(
        pd.concat([all_buildings_gdf, scenario_buildings_gdf], ignore_index=True),
        crs=scenario_buildings_gdf.crs,
        geometry="geom",
    )
```

On the `status2024` iteration the seed set is appended a second time. `eGon2035` takes the
`else` branch, where `allocate_scenarios()` builds a fresh set that was never used as the seed,
which is why only status-quo scenarios are affected.

The bug can only fire when a status-quo scenario is in `SCENARIOS`, so runs predating status-quo
scenarios never hit it.

## Downstream impact

- **`storages/home_batteries.py:127-139`** sums `capacity` per `bus_id` from this table to size
  home batteries. For status-quo scenarios that sum is doubled, so battery capacities derived
  from it are wrong.
- **`sanity_checks.py:681` (`sanitycheck_pv_rooftop_buildings`)** reads the whole table with
  `SELECT *`. Any per-scenario totals or building-level comparisons it makes are affected.
- **`heat_supply/individual_heating.py` (`determine_buildings_with_hp_in_mv_grid`)** queries this
  table to weight buildings with PV at 1.5x when distributing heat pumps. The duplicated rows put
  duplicate entries into the `weights` index, so `np.random.choice` could select the same building
  twice; both copies were then allocated capacity and the second was silently dropped by
  `drop_duplicates("building_id")` in the bulk export, losing that capacity from the grid's budget
  with no rescaling. This caused a ~0.3-0.5 % conservation shortfall in
  `demand.egon_hp_capacity_buildings` for status2024 (~1.68 MW of 349.42 MW in one run) and made
  results non-reproducible between runs. A defensive dedup has been added on the heat-pump side,
  but the underlying data is still wrong for every other consumer.

## Suggested fix

Either seed the accumulator empty and let the loop populate it:

```python
all_buildings_gdf = gpd.GeoDataFrame(
    columns=scenario_buildings_gdf_sq.columns,
    crs=scenario_buildings_gdf_sq.crs,
    geometry="geom",
)
```

or keep the seed and skip the redundant append on the status-quo iteration.

A `UNIQUE` constraint on `(scenario, gens_id)` — or on `(scenario, building_id, gens_id)` — would
turn a recurrence into a hard failure rather than a silent doubling.

## Reproducing

Run `electricity_supply.power_plants.pv_rooftop_buildings.pv-rooftop-to-buildings` with any
status-quo scenario configured, then:

```sql
SELECT scenario, count(*) AS rows, count(DISTINCT building_id) AS distinct_buildings
FROM supply.egon_power_plants_pv_roof_building
GROUP BY 1;
```

Status-quo scenarios show `rows = 2 x distinct_buildings`.

LennartB-dot:
Fixes #1489.

Skips the loop for statusquo scenarios so the data for statusquo is not added a second time.
Checked for SH-runs. 

### 📋 Pull Request Guidelines

---

## 🧑‍💻 Contributor Checklist

- [ ] Relevant **documentation is updated** (API, new features, etc.)
- [ ] Dataset-versions are updated when existing datasets are adjusted.
- [ ] Added a note to `CHANGELOG.rst` about the changes
- [ ] Added yourself to `AUTHORS.rst`

Optional:

- [ ] Extend the checklist for reviewers: Which aspects should be reviewed in particular?

```markdown
<!-- Example:
Please focus on validating the data handling in file XYZ.
-->
```

---

## 🔍 Reviewer Checklist

---

---

💡 **Tip:** If you add multiple reviewers, clarify who should check what — this saves time and avoids duplicated efforts.

### DIFF
--- a/src/egon/data/datasets/power_plants/__init__.py
+++ b/src/egon/data/datasets/power_plants/__init__.py
@@ -1510,7 +1510,7 @@ class PowerPlants(Dataset):
     #:
     name: str = "PowerPlants"
     #:
-    version: str = "0.0.37"
+    version: str = "0.0.38"
 
     def __init__(self, dependencies):
         super().__init__(
--- a/src/egon/data/datasets/power_plants/pv_rooftop_buildings.py
+++ b/src/egon/data/datasets/power_plants/pv_rooftop_buildings.py
@@ -2439,7 +2439,11 @@ def pv_rooftop_to_buildings():
 
     for scenario in SCENARIOS:
         if scenario == status_quo:
-            scenario_buildings_gdf = scenario_buildings_gdf_sq.copy()
+            # status-quo rows are already seeded into all_buildings_gdf above;
+            # other scenarios' allocate_scenarios() calls still rely on
+            # scenario_buildings_gdf_sq / desagg_mastr_gdf as an input, so
+            # those variables are kept intact
+            continue
         elif "status" in scenario:
             # <--- REFACTORING: Use sources.files
             ts = pd.Timestamp(


jh-RLI:
One side effect of the `continue` a bit further down in the same loop: the `elif "status" in scenario` branch builds its mask from `scenario_buildings_gdf`, and that name used to be assigned in the status-quo iteration. With the `continue` it never is, so a run configured with a second `status*` scenario would reference it before assignment.

Not reachable while only one status scenario is in `--scenarios`, so it does not block this fix. Reading `scenario_buildings_gdf_sq.commissioning_date` there looks like the intended frame anyway, and it would make that branch independent of the loop order.

LennartB-dot:
fixed the problem, that the second status-scenario would reference a not defined dataframe.

Issue #1507, opened 2026-08-31
[FEATURE] Implement large scale batteries capacities from NEP into future scenarios


lenzim97:
**Description**
Came up during solving #1478. 
For the new future scenarios (2037/2045) there are future capacities regarding large scale Batteries given in the NEP. 
In Former times future battery capacities were always completely optimized in etrago without any constraints. 
For the future it would be nice to at least have the option to consider the assumptions from the BNetzA for our Optimization. 

**ToDos**
- Decide how to distribute the given target capacities among the federal states (Method decision, probably inside egon-data)
- Imlement this method for defining capcaities for each scenario for each MV/HV-Bus
- Bring optional usage of this values into eTraGo with defining constraints depending on the args

**Current status**
Status capacities coming from MaStR are implemented for status2024 and transfered to future scenarios. Any additional future capacities is part of the optimization in eTraGo.

**Notes**
Reasoning why we don't use target values at the moment:
Capacities coming from NEP are just based on "Netzanschlussbegehren" (25 for scenario B, 50% for scenario C), which are really unsure . BNetzA also want to distribute battery capacities "grid friendly" which is core competence of eTraGo. So we leave it currently as a interesting base for compare our results

Issue #1478, opened 2026-08-10
Distinguishing big scale batteries from Implemenation of batteries


lenzim97:
**Describe the bug**  
Currently there is just an "battery" carrier implemented in the egon_scenario_capacities table. These target values are used to implement and distribute the home_batteries into the database. 
Since there are now also big scale storages in the system, we need to include them to the egon_scneario_capacties and distinguish them from the home_batteries. 

---

**To Dos**  
1. Reserach where current target values for batteries are taken from
2. Secure that for home_batteries only target values are used, that are meant for home_batteries
3. check if there is already something regarding bis scale batteries in the current code
4. Think about how to implement big scale batteries (status, future)

**Environment:**
- OS: [e.g. Windows 11, Ubuntu 22.04]
- Python version: [e.g. Python 3.10, Python 3.11]
- Scenario: [e.g. eGon2035, eGon100RE]
- Dataset Boundary: [Everything or Schleswig-Holstein]

CarlosEpia:
Issue resolved with PR #1509

PR #1509, opened 2026-09-01
Feature/#1478 implementing bess into database

lenzim97:
Solves #1478  .

**Description.**
Distinguish grid-scale battery storage (new carrier 'BESS') from home batteries
throughout the pipeline: real MaStR-based BESS carry-forward (aged trhough all scenarios) 
and separate eTraGo carriers/cost parameters (BESS' vs'home_battery') instead of one generic 'battery' carrier. 
Future capacities from NEP are not considered currently and will be tackled via #1507 as a 'Nice to Have'

### 📋 Pull Request Guidelines

---

**Notes**
- [x] Changes have been tested in **SH_Testmode**

---

## 🔍 Reviewer Checklist

### DIFF
--- a/src/egon/data/datasets/storages/__init__.py
+++ b/src/egon/data/datasets/storages/__init__.py
@@ -117,7 +117,7 @@ class Storages(Dataset):
     #:
     name: str = "Storages"
     #:
-    version: str = "0.0.13"
+    version: str = "0.0.14"
 
     def __init__(self, dependencies):
         super().__init__(
--- a/src/egon/data/datasets/storages/__init__.py
+++ b/src/egon/data/datasets/storages/__init__.py
@@ -117,7 +117,7 @@ class Storages(Dataset):
     #:
     name: str = "Storages"
     #:
-    version: str = "0.0.12"
+    version: str = "0.0.13"
 
     def __init__(self, dependencies):
         super().__init__(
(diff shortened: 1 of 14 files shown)


src/egon/data/datasets/storages/__init__.py
```diff
@@ -722,6 +769,108 @@ def home_batteries_per_scenario(scenario):
...
+
+    mastr["commissioning_date"] = pd.to_datetime(mastr["commissioning_date"], errors="coerce")
+    mastr.loc[mastr["commissioning_date"] < "1990-01-01", "commissioning_date"] = pd.NaT
+    decommissioning_date = pd.to_datetime(mastr["decommissioning_date"], errors="coerce")
+
+    # keep only units already commissioned and not (yet) decommissioned
+    # as of the scenario's reference date
+    mastr = mastr.loc[
+        (mastr["commissioning_date"] < scenario_date_max)
+        & (decommissioning_date.isna() | (decommissioning_date > scenario_date_max))
```
CarlosEpia, line 796: I don't know if this condition is a bit too restrictive:
decommissioning_date.isna()

Do you know how much capacity are we dropping by using that piece of filter?

CarlosEpia (approved):
If the total dropped battery capacity is not too high (see comment) this PR can be merged. If that is not the case we need to find a way to make it less restrictive, even deleting that constraint is an option.

src/egon/data/datasets/storages/__init__.py
lenzim97: For 'BESS' it does only have any impact at regon2045 cause of the lifespan of the batterie.  For 'home_batteries it has mostly an impact under 1%, only for th status scenario there are more batteries filtered out since a lot of batteries where built after 2024.
Here is the overview: 

Image

PR #1511, opened 2026-09-01
Fixes/#1510 re gon nep2025 scenarios conventional and pumped hydro power plant allocation broken

### LINKED ISSUE #1510: [BUG] reGon (NEP2025) scenarios: conventional & pumped-hydro power plant allocation broken
Running the pipeline with `reGon2037` / `reGon2045` configured **without** `eGon2035`
(e.g. `--scenarios status2024,reGon2037`). The NEP2025 "Kraftwerksliste
Szenariorahmen" is loaded into `supply.egon_nep_conventional_powerplants` and
consumed by the power-plant and storage allocation datasets.

`power_plants.allocate-conventional-non-chp-power-plants` **fails**:

psycopg2.errors.UndefinedColumn: column "bnetza_id" does not exist
[SQL: SELECT bnetza_id, name, carrier, capacity, postcode, city,
federal_state, c2037_capacity FROM supply.egon_nep_conventional_powerplants
WHERE carrier = 'oil' AND scenario = 'reGon' AND chp = 'Nein' ...]

After working around (1), the task **succeeds but allocates 0 MW** for reGon –
both `oil` and `gas` NEP DataFrames come back empty.

`storages.allocate-pumped-hydro-scn` fails with the same `bnetza_id` error on
its own reGon query.

CarlosEpia:
Fixes #1510.

Already tested for Germany. Results in server run.

**Abstract of changes:**

- Make the conventional non-CHP and pumped-hydro power plant allocation work with the NEP2025 Kraftwerksliste (reGon2037/reGon2045), which identifies plants by `mastr_id` instead of `bnetza_id`. the id column is now chosen per NEP list version instead of being hard-coded.
- Normalize the NEP CHP flag to `Ja`/`Nein` on import, so the lowercase `ja`/`nein` of the NEP2025 list no longer slips past the downstream `chp = 'Nein'` filters and silently drops every non-CHP reGon plant.
- Bump the affected dataset versions (`PowerPlants`, `ScenarioCapacities`, `Storages`) so the corrected steps re-run.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -118,6 +118,16 @@ Bug Fixes
   regular pass and inserted a second identical load under the scenario's
   own name, doubling transport demand in status quo scenarios
   `#1483 <https://github.com/openego/eGon-data/issues/1483>`_
+* Fix conventional non-CHP and pumped hydro power plant allocation for the
+  reGon2037/reGon2045 scenarios: the NEP2025 Kraftwerksliste has no
+  ``bnetza_id`` column (only ``mastr_id``) and stores the CHP flag as
+  lowercase ``ja``/``nein``, so ``select_nep_power_plants`` /
+  ``select_nep_pumped_hydro`` crashed with ``UndefinedColumn`` and, once
+  that was worked around, allocated nothing because every ``chp = 'Nein'``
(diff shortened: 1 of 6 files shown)


LennartB-dot (approved):
looking good.
Thanks!

PR #1516, opened 2026-09-09
Add public bus charging dataset (vehicle class M3)

### LINKED ISSUE #1461: [FEATURE] New eMobility dataset for public buses (vehicle class M3)
As required by the new NEP public road transport (BUSES) will be added.

Static timeseries will be part of the new data bundle #1437.
The pipeline will create new static loads to the eTraGo model.

MoritzSchloesser:
Fixes #1461.

Adds the eMobility dataset for public buses (vehicle class M3): depot locations with a
static hourly charging series per scenario. No flexibility is modelled, though it may be
added later.

Two tables are written from one source of truth:

- **`demand.egon_ev_bus_charging_depot`** — one row per (depot, scenario) with geometry,
  `depot_type`, `fleet_size`, `voltage_level`, `peak_load_mw`, `annual_demand_mwh` and
  the full 8760-step `p_set`. This is the per-location table for eDisGo, which cannot
  regenerate a predefined series from an aggregate.
- **`grid.egon_etrago_load` / `_timeseries`** — one load per eTraGo bus per scenario
  under the new carrier `land_transport_bus`, with the depots' series summed.

Unlike MIT and HGV, the demand is not generated inside egon-data: the depot gpkg and
three gzipped hourly CSVs are produced externally and ship in the data bundle under
`data_bundle_egon_data/bus_charging/`. This dataset is the integration layer only.

**Scenario coverage**

| Scenario | Bus data | Depots | Annual demand |
|---|---|---|---|
| `status2024` | yes | 1053 | 0.246 TWh |
| `reGon2037` | yes | 1352 | 3.106 TWh |
| `reGon2045` | yes | 1352 | 4.350 TWh |
| `eGon2035` | no — deliberately unchanged | — | — |
| `eGon100RE` | no — discarded | — | — |

Zero bus demand in `eGon2035` / `eGon100RE` is correct rather than a missing-data bug.
The `status2024` figure is a sanity anchor: 0.246 TWh at 70–84 MWh/bus/a implies
~2,900–3,500 buses, matching the 2,946 battery-electric buses actually registered in
Germany at end-2024 (KBA / PwC E-Bus-Radar) — it models the currently electrified
fraction, not the full fleet.

### 📋 Pull Request Guidelines

---

## 🧑‍💻 Contributor Checklist

      _`pytest -vv --ignore=src`: 3 passed, 1 skipped (the skip needs Docker/PostgreSQL). `flake8` and `isort` pass on the changed files._
- [x] Workflow has run at least once in **Test mode**

Optional:

```markdown
Please focus on:

1. The eTraGo aggregation strategy (`etrago_integration.py`). Depots are summed per
   eTraGo bus rather than written individually. An MV grid district is the Voronoi cell
   of exactly one HV/MV substation, so depots in a district necessarily share a bus, and
   with no flexibility to distinguish them N loads on one bus are mathematically
   identical to one summed load. Sanity-check that reasoning.

2. The per-scenario voltage level (`spatial_assignment.py`). 560 of the 1053 depots
   present in all three scenarios are assigned a higher voltage level in reGon2045 than
   in status2024, and no connection upgrade is modelled. This follows the convention for
   other growing loads, but it is the decision with the widest downstream effect.

3. The two-layer bus assignment. Levels 3–7 join against `grid.egon_mv_grid_district`,
   levels 1–2 against `grid.egon_ehv_substation_voronoi`, following `power_plants`. The
   EHV path is unexercised by the current input data (no depot exceeds 120 MW) and only
   logs a warning — review whether that is acceptable or should raise.

4. Whether `land_transport_bus` should be registered in
   `etrago_setup.insert_carriers()`. HGV's `land_transport_HGV` is not, and its loads
   write fine, so I left it out — but if that list is meant to be exhaustive it is a
   one-line addition.
```

---

## 🔍 Reviewer Checklist

---

### Design decisions

Recorded as ADRs alongside the input data; the two non-obvious ones:

**Voltage level is per (depot, scenario), and connection upgrades are not modelled.**
Of the 1053 depots present in all three scenarios, only 493 keep the same voltage level
from `status2024` to `reGon2045` — 560 are upgraded, none downgraded. This follows the
convention every other growing load uses (industry, heat pumps, PV rooftop, power
plants) and keeps `status2024` physically honest: a depot charging two buses today
genuinely is on an LV connection, and 393 of its 1053 depots are at level 6–7. The
trade-off is that cross-scenario connection-upgrade cost stays invisible — a property of
egon-data's scenario design rather than of this dataset.

**Per-depot series are stored even though eTraGo only needs the aggregate,** because
depot peaks are not coincident: the sum of individual peaks in `reGon2045` is 3770 MW
against a lower aggregate peak. Only the full series preserves *when* each depot peaks,
which is what LV reinforcement analysis needs. A normalized national profile scaled by
annual energy — the approach HGV uses for highway-day charging — would not work here,
because each depot's shape is genuinely distinct. Cost is ~300 MB at national extent.

### Implementation notes

- Out-of-boundary depots are **dropped, not snapped to the nearest bus**. Under
  `--dataset-boundary` the district table covers only the test region, so a
  nearest-match fallback would silently pile every out-of-region depot onto the
  boundary's edge buses — a failure that looks plausible and goes undetected.
- Deletion is scoped by carrier **and** scenario, so a run configured for one scenario
  does not wipe another's bus loads. (MIT filters by carrier and German buses but not by
  scenario, which is a footgun for incremental runs.)
- `depot_id` is a **String** primary key, unlike the integer keys used elsewhere. It is
  the natural key, joins directly to the input CSV column headers, and stays stable
  across input-data regenerations — an integer surrogate would either need a mapping
  table or would silently renumber depots when the input changes.
- `osmtgmod` is a **required** dependency, not redundant with `mv_grid_districts`:
  `substation_voronoi` depends only on the single task
  `electricity_grid.osmtgmod.substation.extract`, which sits in the same parallel task
  set as `to_pypsa`, so its completion does not imply `grid.egon_etrago_bus` is
  populated.
- `p_set` is inserted via the ORM rather than `GeoDataFrame.to_postgis`: the latter
  writes through `COPY`, which renders a Python list as `[...]` and PostgreSQL rejects
  it as a malformed array literal. Worth knowing for anyone adding an `ARRAY` column
  alongside a geometry.

### Known input-data defect

The `reGon2037` and `reGon2045` series each carry **1821 NaN across 34 depots**, in 5
distinct patterns. Every one sits at hour 0, 1 or 2 **on a Sunday** — 96 rows over 32
distinct Sundays — the signature of a service-day boundary in the source timetable join
rather than noise. The `status2024` file is clean.

Pending a fix upstream these are zero-filled so the pipeline can run. This understates
those depots' demand (~0.6 % of hours for an affected depot, in a low-demand window), so
the substitution is logged with the affected depot ids on every run and reported
separately in the validation notebook. It should be removed and the error restored once
the input data is corrected.

### Testing

Verified by a full Airflow DAG run (`manual__2026-09-08T14:30:25`) in **Test mode**
(`--dataset-boundary: Schleswig-Holstein`, all four scenarios configured) — all four
tasks SUCCESS:

```
Read 1352 depots from depots.gpkg
  status2024: wrote 1053 depots, 0.246 TWh, peak sum 413.9 MW
  reGon2037: zero-filling 1821 missing values across 34 depots
  reGon2037: wrote 1352 depots, 3.106 TWh, peak sum 2771.1 MW
  reGon2045: wrote 1352 depots, 4.350 TWh, peak sum 3770.2 MW
201 MV grid districts, 26 EHV Voronoi cells
  → 65 depots assigned to 62 buses per scenario
  → 62 eTraGo loads per scenario
```

Not yet run in **Everything mode**.

A validation notebook (kept with the input data, outside this repo) runs 11 checks in
two modes — against the raw input files and against the pipeline output. All pass. Five
compare the database against the input rather than against itself:

| # | Check | Result |
|---|---|---|
| 7 | Depot set is a subset of the input | 0 invented rows |
| 8 | `peak_load_mw` / `annual_demand_mwh` / `voltage_level` match input | 0 mismatches |
| 9 | `p_set` matches the input **element-wise** | 0 of 1,708,200 values differ |
| 10 | Each scenario holds its **own** input file | zero diagonal, non-zero off-diagonal |
| 11 | gpkg attributes and geometry preserved | 0 mismatches, 0.0000 m geometry shift |

Check 10 matters because `reGon2037` and `reGon2045` differ by only 1.97 MW at maximum,
so a mix-up between those two needs an element-wise test rather than an aggregate one.
Energy conservation through the eTraGo aggregation holds to floating-point precision
(max relative difference 4.6e-16).

### Open points

- **The data bundle needs republishing.** The current Zenodo bundle
  (`record/16576506`) does not yet contain `bus_charging/`. Local and test runs work
  because the files are staged in the working directory, but a fresh run or CI will fail
  at `fill_depot_table` with an explicit `FileNotFoundError` until the bundle is updated.
- **No OEP metadata yet** for `demand.egon_ev_bus_charging_depot` — the other eMobility
  datasets carry an `add_metadata()` task. Happy to add it here, or file a follow-up
  issue if that is the usual route.
- **No entry in `sanity_checks.py`.** That module registers tasks only for `eGon2035`
  and `eGon100RE`, so under the default `--scenarios` the task tuple is empty and
  nothing would run. Validation currently lives in the notebook.
- **No en-route (opportunity) charging.** The input assumes all M3 charging happens at
  depots; energy a real network would draw at route termini is folded into the depot
  totals. Annual energy is complete, but daytime terminus demand appears as overnight
  depot demand. Adding it would need new input data, not just a schema change.
- **eDisGo handoff not yet confirmed.** This dataset supplies `peak_load_mw`, but peak
  load and contracted grid connection capacity are not the same thing, and whether
  eDisGo can ingest an LV-connected load with a predefined time series (393 of 1053
  `status2024` depots are at level 6–7) still needs checking.

### Incidental finding (not addressed here)

`industry/temporal.py:115-121` names a variable `ehv_voronoi` and comments "Select ehv
voronoi", but the SQL queries `grid.egon_mv_grid_district`. The intent was clearly the
EHV table. It is never hit in practice because no industrial load exceeds 120 MW, so it
is latent rather than an active bug — worth a separate issue.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -2,4 +2,4 @@
 Authors
 =======
 
-* Jonathan Amme, Clara Büttner, Ilka Cußmann, Julian Endres, Carlos Epia, Kilian Helfenbein, Stephan Günther, Ulf Müller, Amélia Nadal, Guido Pleßmann, Francesco Witte  - https://github.com/openego/eGon-data
+* Jonathan Amme, Clara Büttner, Ilka Cußmann, Julian Endres, Carlos Epia, Kilian Helfenbein, Stephan Günther, Ulf Müller, Amélia Nadal, Guido Pleßmann, Moritz Schlösser, Francesco Witte  - https://github.com/openego/eGon-data
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,6 +19,11 @@ Added
   alongside home batteries, and separate eTraGo carriers/cost parameters 
   ('BESS' vs'home_battery') instead of one generic 'battery' carrier
   `#1478 <https://github.com/openego/eGon-data/issues/1478>`_
+* Add new eMobility dataset for public buses (vehicle class M3): static depot
+  charging loads per scenario, written per-depot to
+  demand.egon_ev_bus_charging_depot for eDisGo and aggregated per eTraGo bus
+  under the new carrier 'land_transport_bus'
+  `#1461 <https://github.com/openego/eGon-data/issues/1461>`_
 
 
 Changed
--- a/docs/data/mobility_demand.rst
+++ b/docs/data/mobility_demand.rst
@@ -179,3 +179,55 @@ of each NUTS3 region is used to determine the respective hydrogen Voronoi cell (
 :py:class:`GasAreaseGon2035<egon.data.datasets.gas_areas.GasAreaseGon2035>` and
 :py:class:`GasAreaseGon100RE<egon.data.datasets.gas_areas.GasAreaseGon100RE>`) it is
 located in.
+
+.. _mobility-demand-public-bus-ref:
+
+Public buses
+++++++++++++
+
+Electricity demand of public buses (vehicle class M3) is set up for the
+``status2024``, ``reGon2037`` and ``reGon2045`` scenarios by
+:py:class:`PublicBusCharging<egon.data.datasets.emobility.public_bus_charging.PublicBusCharging>`.
+The ``eGon2035`` and ``eGon100RE`` scenarios deliberately carry no public bus
+demand.
+
+Unlike motorized individual travel and heavy-duty transport, the charging
+demand is not generated within egon-data. Depot locations and a static hourly
+charging series per depot and scenario are produced externally and ship in the
+data bundle under ``data_bundle_egon_data/bus_charging/``. No flexibility is
+modelled: the series are taken as given, so no charging link or storage
+component is written and no lowflex scenario variant is created.
+
+Each depot is written per scenario to
+:py:class:`demand.egon_ev_bus_charging_depot<egon.data.datasets.emobility.public_bus_charging.db_classes.EgonEvBusChargingDepot>`
+with its geometry, grid connection voltage level and full 8760-step time
+series. This per-location detail is required by eDisGo, which cannot
+reconstruct a predefined series from an aggregate.
+
+The voltage level is derived per depot **and per scenario** from that
+scenario's own peak load, using the same thresholds as
+:py:func:`identify_voltage_level<egon.data.datasets.industry.temporal.identify_voltage_level>`.
+Because bus electrification grows steeply between the scenarios, the majority
+of depots cross a threshold: only 493 of the 1053 depots present in all three
+scenarios keep the same voltage level, and 560 are assigned a higher one.
+Consistent with the treatment of other growing loads in egon-data, connection
+upgrades between scenarios are not modelled as events.
+
+The grid connection point is found by a spatial join: depots at voltage levels
+3 to 7 against ``grid.egon_mv_grid_district``, and those at levels 1 or 2
+against ``grid.egon_ehv_substation_voronoi``. Since an MV grid district is the
+Voronoi cell of a single HV/MV substation, all depots within one district share
+that district's eTraGo bus. Depots that fall into no district are dropped
+rather than matched to the nearest one.
+
+For eTraGo, the depot series are summed per bus and written to
+``grid.egon_etrago_load`` and ``grid.egon_etrago_load_timeseries`` under the
+carrier ``land_transport_bus``.
+
+.. note::
+   The ``reGon2037`` and ``reGon2045`` input series each contain 1821 missing
+   values across 34 depots, all falling in the first three hours of a Sunday.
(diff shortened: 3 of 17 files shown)


nesnoj:
Quickly spotted:
- Why are there more depots in 2037/2045 compared to `status2024`?
- The docs have not been updated

